PR #3, opened 2021-02-21
[repomanager] Initial contribution

J-N-K:
Initial contribution of a repository manager


J-N-K:
@splatch I would be very glad if you could have a look. This is a very basic version to get things started. Later versions could use configuration instead of hard-coded URLs and then support also other repositories.

splatch:
@J-N-K I had a look this morning and it looks pretty good for first incarnation. Overall repository management with `pax-url-mvn` is rather tough topic. I remember my failed fights few years ago with unification of version range resolution and authorization injection. Way you did is fairly reasonable.

The only instant suggestion I  have right now is usage of `OpenHabServlet` as a base for repository servlet you made. Just toprotect it with security layer. I think OHC currently might not have a default security handlers in place so it doesn't change much, but leaves space for plugging auth if necessary.

Because pax-url shades maven repository management the actual logic is hidden behind facades - `org.ops4j.pax.url.mvn.MavenResolver` and `org.ops4j.pax.url.mvn.internal.config.MavenConfiguration`. If you will have a look there you will find that there is an API to submit online queries. :-) If needed I think we will be able to drive pax-url a bit forward with support for storing modified file (I have commit karma there).

PR #20, opened 2021-03-02
[infrastructure] upgrade bnd and improve build

J-N-K:

### DIFF
--- a/bom/smarthomej-addons/pom.xml
+++ b/bom/smarthomej-addons/pom.xml
@@ -43,6 +43,11 @@
     </dependency>
     <dependency>
       <groupId>org.smarthomej.addons.bundles</groupId>
+      <artifactId>org.smarthomej.binding.knx</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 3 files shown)


splatch:
A small suggestion for bnd and headers in general - if possible try including build date. I know it is getting there via qualifier or another header but usually it is ill formatted or hard to read for end user. Macro I used for before is `${now;dd.MM.YYYY HH:mm}`. It will be easier for people who rely on snapshots to identify what they got.

J-N-K:
Good point. The question is which header is visible to the user and NOT restricted in format?

splatch:
> The question is which header is visible to the user and NOT restricted in format?

There are just few headers which come from osgi spec such as `Bundle-Version`. You can pick any other, ie. `Build-Datetime`.
I dont think there is any tool which will parse it. 
I am not familiar with travis, but maven semantics force you to deploy jars with snapshot suffix which is where osgi qualifier comes in.

User will find custom header with `bundle:headers` command in the shell. I know its rather counter intuitive but so far there is no other way to list bundle headers and custom meta.

PR #43, opened 2021-03-21
[androiddebugbridge] Timeout fix and code improvement

Trinitus01:
changed: clean up code
changed: exception messages
changed: startCommand -> monkey to am start
fixed: wake-lock because of line breaks
fixed: "Unable to refresh awake state: Timeout"

### DIFF
--- a/bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
+++ b/bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
@@ -18,6 +18,7 @@ import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBri
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.MEDIA_CONTROL_CHANNEL;
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.MEDIA_VOLUME_CHANNEL;
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.SCREEN_STATE_CHANNEL;
+import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.WAKE_LOCK_CHANNEL;
 
 import java.io.*;
 import java.net.InetSocketAddress;
@@ -29,6 +30,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Arrays;
 import java.util.Base64;
+import java.util.Optional;
 import java.util.concurrent.*;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -62,7 +64,7 @@ public class AndroidDebugBridgeDevice {
     private static @Nullable AdbCrypto adbCrypto;
 
     static {
-        var logger = LoggerFactory.getLogger(AndroidDebugBridgeDevice.class);
+        Logger logger = LoggerFactory.getLogger(AndroidDebugBridgeDevice.class);
         try {
             File directory = new File(ADB_FOLDER);
             if (!directory.exists()) {
@@ -118,70 +120,67 @@ public class AndroidDebugBridgeDevice {
 
     public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, InterruptedException,
             AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
-        var out = runAdbShell("dumpsys", "window", "windows", "|", "grep", "mFocusedApp");
-        var targetLine = Arrays.stream(out.split("\n")).findFirst().orElse("");
-        var lineParts = targetLine.split(" ");
+        String result = runAdbShell("dumpsys", "window", "windows", "|", "grep", "mFocusedApp");
+        String targetLine = Arrays.stream(result.split("\n")).findFirst().orElse("");
+        String[] lineParts = targetLine.split(" ");
         if (lineParts.length >= 2) {
-            var packageActivityName = lineParts[lineParts.length - 2];
+            String packageActivityName = lineParts[lineParts.length - 2];
             if (packageActivityName.contains("/"))
                 return packageActivityName.split("/")[0];
         }
-        throw new AndroidDebugBridgeDeviceReadException(
-                "Device does not support " + CURRENT_PACKAGE_CHANNEL + ": " + out);
+        throw new AndroidDebugBridgeDeviceReadException(CURRENT_PACKAGE_CHANNEL, result);
     }
 
-    public @Nullable Boolean isAwake() throws InterruptedException, AndroidDebugBridgeDeviceException,
+    public Optional<Boolean> isAwake() throws InterruptedException, AndroidDebugBridgeDeviceException,
             AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
-        var out = runAdbShell("dumpsys", "activity", "|", "grep", "mWakefulness");
-        if (out.contains("mWakefulness=")) {
-            return out.contains("mWakefulness=Awake");
-        } else if (out.isEmpty()) {
-            return null;
+        String result = runAdbShell("dumpsys", "activity", "|", "grep", "mWakefulness");
+        if (result.contains("mWakefulness=")) {
+            return Optional.of(result.contains("mWakefulness=Awake"));
+        } else if (result.isEmpty()) {
+            return Optional.empty();
         }
-        throw new AndroidDebugBridgeDeviceReadException("Device does not support " + AWAKE_STATE_CHANNEL + ": " + out);
+        throw new AndroidDebugBridgeDeviceReadException(AWAKE_STATE_CHANNEL, result);
     }
 
-    public @Nullable Boolean isScreenOn() throws InterruptedException, AndroidDebugBridgeDeviceException,
+    public Optional<Boolean> isScreenOn() throws InterruptedException, AndroidDebugBridgeDeviceException,
             AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
-        var out = runAdbShell("dumpsys", "power", "|", "grep", "'Display Power'");
-        if (out.contains("=")) {
-            var state = out.split("=")[1];
-            return state.equals("ON");
-        } else if (out.isEmpty()) {
-            return null;
+        String result = runAdbShell("dumpsys", "power", "|", "grep", "'Display Power'");
+        String[] splitResult = result.split("=");
+        if (splitResult.length >= 2) {
+            return Optional.of("ON".equals(splitResult[1]));
+        } else if (result.isEmpty()) {
+            return Optional.empty();
         }
-        throw new AndroidDebugBridgeDeviceReadException("Device does not support " + SCREEN_STATE_CHANNEL + ": " + out);
+        throw new AndroidDebugBridgeDeviceReadException(SCREEN_STATE_CHANNEL, result);
     }
 
-    public @Nullable Boolean isPlayingMedia(String currentApp) throws AndroidDebugBridgeDeviceException,
+    public Optional<Boolean> isPlayingMedia(String currentApp) throws AndroidDebugBridgeDeviceException,
             AndroidDebugBridgeDeviceReadException, InterruptedException, TimeoutException, ExecutionException {
-        var out = runAdbShell("dumpsys", "media_session", "|", "grep", "-A", "100", "'Sessions Stack'", "|", "grep",
-                "-A", "50", currentApp);
-        String[] mediaSessions = out.split("\n\n");
+        String result = runAdbShell("dumpsys", "media_session", "|", "grep", "-A", "100", "'Sessions Stack'", "|",
+                "grep", "-A", "50", currentApp);
+        String[] mediaSessions = result.split("\n\n");
         if (mediaSessions.length == 0) {
             // no media session found for current app
-            return false;
+            return Optional.of(false);
         } else if (mediaSessions[0].contains("PlaybackState {state=3")) {
             boolean isPlaying = mediaSessions[0].contains("PlaybackState {state=3");
             logger.debug("device media state playing {}", isPlaying);
-            return isPlaying;
-        } else if (out.isEmpty()) {
-            return null;
+            return Optional.of(isPlaying);
+        } else if (result.isEmpty()) {
+            return Optional.empty();
         }
-        throw new AndroidDebugBridgeDeviceReadException(
-                "Device does not support " + MEDIA_CONTROL_CHANNEL + ": " + out);
+        throw new AndroidDebugBridgeDeviceReadException(MEDIA_CONTROL_CHANNEL, result);
     }
 
-    public @Nullable Boolean isPlayingAudio() throws AndroidDebugBridgeDeviceException,
+    public Optional<Boolean> isPlayingAudio() throws AndroidDebugBridgeDeviceException,
             AndroidDebugBridgeDeviceReadException, InterruptedException, TimeoutException, ExecutionException {
-        var out = runAdbShell("dumpsys", "audio", "|", "grep", "ID:");
-        if (out.contains("state:")) {
-            return out.contains("state:started");
-        } else if (out.isEmpty()) {
-            return null;
+        String result = runAdbShell("dumpsys", "audio", "|", "grep", "ID:");
+        if (result.contains("state:")) {
+            return Optional.of(result.contains("state:started"));
+        } else if (result.isEmpty()) {
+            return Optional.empty();
         }
-        throw new AndroidDebugBridgeDeviceReadException(
-                "Device does not support " + MEDIA_CONTROL_CHANNEL + ": " + out);
+        throw new AndroidDebugBridgeDeviceReadException(MEDIA_CONTROL_CHANNEL, result);
     }
 
     public VolumeInfo getMediaVolume() throws AndroidDebugBridgeDeviceException, InterruptedException,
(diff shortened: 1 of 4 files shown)


bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -118,47 +126,62 @@ public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, Inte
...
-        throw new AndroidDebugBridgeDeviceReadException("Unable to read package name");
+        throw new AndroidDebugBridgeDeviceReadException(
+                "Device does not support " + CURRENT_PACKAGE_CHANNEL + ": " + out);
     }
 
-    public boolean isAwake()
-            throws InterruptedException, AndroidDebugBridgeDeviceException, TimeoutException, ExecutionException {
-        String devicesResp = runAdbShell("dumpsys", "activity", "|", "grep", "mWakefulness");
-        return devicesResp.contains("mWakefulness=Awake");
+    public @Nullable Boolean isAwake() throws InterruptedException, AndroidDebugBridgeDeviceException,
```
J-N-K, line 133:
```suggestion
    public Optional<Boolean> isAwake() throws InterruptedException, AndroidDebugBridgeDeviceException,
```

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -118,47 +126,62 @@ public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, Inte
...
+            AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
+        var out = runAdbShell("dumpsys", "activity", "|", "grep", "mWakefulness");
+        if (out.contains("mWakefulness=")) {
+            return out.contains("mWakefulness=Awake");
```
J-N-K, line 137:
```suggestion
            return Optional.of(out.contains("mWakefulness=Awake"));
```

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -118,47 +126,62 @@ public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, Inte
...
+        } else if (out.isEmpty()) {
+            return null;
```
J-N-K, line 139:
```suggestion
            return Optional.empty()
```

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -118,47 +126,62 @@ public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, Inte
...
-        if (devicesResp.contains("=")) {
-            try {
-                var state = devicesResp.split("=")[1].trim();
-                return state.equals("ON");
-            } catch (NumberFormatException e) {
-                logger.debug("Unable to parse device wake lock: {}", e.getMessage());
-            }
+        var out = runAdbShell("dumpsys", "power", "|", "grep", "'Display Power'");
+        if (out.contains("=")) {
+            var state = out.split("=")[1];
```
J-N-K, line 148: See below: potentioan `IndexOutOfBoundsException`

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -273,25 +299,32 @@ private String runAdbShell(String... args)
...
-                do {
-                    byteArrayOutputStream.writeBytes(stream.read());
-                } while (!stream.isClosed());
-            } catch (IOException e) {
-                String message = e.getMessage();
-                if (message != null && !message.equals("Stream closed")) {
-                    throw e;
+        lock.lock();
+        try {
+            commandFuture = scheduler.submit(() -> {
```
J-N-K, line 304: Is this still needed as a field in the class if there is a lock?

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -254,16 +278,18 @@ public void connect() throws AndroidDebugBridgeDeviceException, InterruptedExcep
             if (e.getMessage().equals("Socket closed")) {
                 // Connection aborted by us
                 throw new InterruptedException();
+            } else if (e.getMessage().equals("No route to host (Host unreachable)")) {
+                throw new TimeoutException();
```
J-N-K, line 282: Are these exceptions treted differently? Otherwise it's usually better to stick with one exception type.

J-N-K (changes requested):

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -118,47 +126,62 @@ public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, Inte
...
+                "Device does not support " + CURRENT_PACKAGE_CHANNEL + ": " + out);
     }
 
-    public boolean isAwake()
-            throws InterruptedException, AndroidDebugBridgeDeviceException, TimeoutException, ExecutionException {
-        String devicesResp = runAdbShell("dumpsys", "activity", "|", "grep", "mWakefulness");
-        return devicesResp.contains("mWakefulness=Awake");
+    public @Nullable Boolean isAwake() throws InterruptedException, AndroidDebugBridgeDeviceException,
+            AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
+        var out = runAdbShell("dumpsys", "activity", "|", "grep", "mWakefulness");
```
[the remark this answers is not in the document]
Trinitus01: fixed

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -118,47 +126,62 @@ public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, Inte
...
+            logger.debug("device media state playing {}", isPlaying);
+            return isPlaying;
+        } else if (out.isEmpty()) {
+            return null;
         }
-        boolean isPlaying = mediaSessions[0].contains("PlaybackState {state=3");
-        logger.debug("device media state playing {}", isPlaying);
-        return isPlaying;
+        throw new AndroidDebugBridgeDeviceReadException(
+                "Device does not support " + MEDIA_CONTROL_CHANNEL + ": " + out);
```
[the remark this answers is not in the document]
Trinitus01: fixed, super() must be the first line so unable to use StringBuilder instead of string + string + string + aso

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -173,15 +196,15 @@ public void setMediaVolume(int volume)
 
     public int getPowerWakeLock() throws InterruptedException, AndroidDebugBridgeDeviceException,
             AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
-        String lockResp = runAdbShell("dumpsys", "power", "|", "grep", "Locks", "|", "grep", "'size='");
-        if (lockResp.contains("=")) {
+        var out = runAdbShell("dumpsys", "power", "|", "grep", "Locks", "|", "grep", "'size='");
+        if (out.contains("=")) {
             try {
-                return Integer.parseInt(lockResp.replace("\n", "").split("=")[1].trim());
+                return Integer.parseInt(out.split("=")[1]);
```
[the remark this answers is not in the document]
Trinitus01: fixed

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -254,16 +278,18 @@ public void connect() throws AndroidDebugBridgeDeviceException, InterruptedExcep
             if (e.getMessage().equals("Socket closed")) {
                 // Connection aborted by us
                 throw new InterruptedException();
+            } else if (e.getMessage().equals("No route to host (Host unreachable)")) {
```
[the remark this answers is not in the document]
Trinitus01: fixed

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -254,16 +278,18 @@ public void connect() throws AndroidDebugBridgeDeviceException, InterruptedExcep
             if (e.getMessage().equals("Socket closed")) {
```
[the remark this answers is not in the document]
Trinitus01: fixed

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
```diff
@@ -139,14 +143,18 @@ private void handleCommandInternal(ChannelUID channelUID, Command command)
                 break;
             case AWAKE_STATE_CHANNEL:
                 if (command instanceof RefreshType) {
-                    boolean awakeState = adbConnection.isAwake();
-                    updateState(channelUID, OnOffType.from(awakeState));
+                    Boolean awakeState = adbConnection.isAwake();
+                    if (awakeState != null) {
+                        updateState(channelUID, OnOffType.from(awakeState));
+                    }
```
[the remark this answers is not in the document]
Trinitus01: fixed, but do not understand why to make it more complicated with optional instead to use null as a status

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Trinitus01: did but do not understand why to make it more complicated with optional instead to use null as a status

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Trinitus01: did but do not understand why to make it more complicated with optional instead to use null as a status

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Trinitus01: did but do not understand why to make it more complicated with optional instead to use null as a status

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Trinitus01: fixed

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Trinitus01: yes because the disconnect method needs to cancel it as soon at it is called. if the thread timeout for example in 30 seconds and the disconnect method is called in the 7th second it has to clean up the thread now. else the thread will be alive for 23 seconds more and this could maybe produce a schedule overflow.

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Trinitus01: yes because in the upper method it use the exception to decide what to show on the frontend: offline or err:communication

Trinitus01:
After the changes all the methods with return of "Optional" do not work anymore, sadly.

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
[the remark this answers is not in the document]
J-N-K: Using `Optional`prevents NPE. If you return `null` you need to check if the return-value is null (and this could be forgotten). Another option would be to return the primitive `boolean` and thrwo a checked exception (e.g. `ValueNotFoundException` or something names similar) and catch that. But usually the `Optional` is easier and requires less code.

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
J-N-K: Woudl it make sense to check if a `commandFuture` is already set (check for null) and set it to `null` in the `finally` clause?

J-N-K (changes requested):
Otherwise looks good.

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Trinitus01: Additional to the disconnect method or as a replacement for it? I think after a cancel a future should always set to null but in the aec-binding we also only set it on the logout method to null, so i did it the same way in the adb-binding. I still don't know if there is any difference but i hope the garbage collector will collect it in any way (is null or thread finished/canceled) I only can say that if an timeout exception appears in the thread because the get method timed out, the thread is still alive and is in a strange condition. If this happens multiple times it seems the max threads of scheduler is reached and no command/thread is executed anymore. This is why we have to cancel it in the finally block, prophylactically. This will stop the scheduler-overflow (same i recognized on the aec-binding before you changed it like it is today). And in the disconnect method like i explained before.

J-N-K: The garbage collector will collect it if there is no reference to the object, so it'll be collected. But If the timer was not finished before it is replaced, the timer will execute at the time set before. So generally before assigning a new value it is a good idea to check if there is already a scheduled job.  I would recommend a method like this:

```
private void stopCommandFuture() {
    Future<?> f = commandFuture;
    if (f != null) {
        f.cancel(true);
        commandFuture = null;
    }
 }
```

and call this in all places where you want to end the current job (e.g. in the `finally` clause instead of l. 322-324 and in `disconnect` instead of l. 356-360) and also in all places where you assign a new instance to the `commandFuture` field (before l. 304) . This ensures that at no time there is more than one command future in place.

Regarding the amazonechocontrol: There is no need to set it to `null`, because (e.g. the annoucnement-timer) removed from the `timers`-map here: 

https://github.com/smarthomej/addons/blob/28d6b1f6fff9acfd3749e3634fe369e345cc061c/bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/Connection.java#L1365 

That is essentially the same as setting it to null, because further `get` calls to the map will result in `null`.  New values are only assigned if there is no job currently scheduled:

https://github.com/smarthomej/addons/blob/28d6b1f6fff9acfd3749e3634fe369e345cc061c/bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/Connection.java#L1326-L1327

J-N-K (approved):

PR #48, opened 2021-03-25
[androiddebugbridge] Missing part to refresh hdmi

Trinitus01:
added: missing part to refresh hdmi

### DIFF
--- a/bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
+++ b/bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
@@ -19,6 +19,7 @@ import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBri
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.MEDIA_CONTROL_CHANNEL;
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.MEDIA_VOLUME_CHANNEL;
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.SCREEN_STATE_CHANNEL;
+import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.START_PACKAGE_CHANNEL;
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.WAKE_LOCK_CHANNEL;
 
 import java.io.*;
@@ -31,6 +32,8 @@ import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Arrays;
 import java.util.Base64;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.*;
 import java.util.concurrent.locks.Lock;
@@ -88,6 +91,7 @@ public class AndroidDebugBridgeDevice {
     private @Nullable Future<String> commandFuture;
 
     private Lock lock = new ReentrantLock();
+    private Map<String, @Nullable Boolean> fallbacks = new HashMap<>();
 
     AndroidDebugBridgeDevice(ScheduledExecutorService scheduler) {
         this.scheduler = scheduler;
@@ -117,9 +121,18 @@ public class AndroidDebugBridgeDevice {
         runAdbShell("input", "text", URLEncoder.encode(text, StandardCharsets.UTF_8));
     }
 
-    public void startPackage(String packageName)
-            throws InterruptedException, AndroidDebugBridgeDeviceException, TimeoutException, ExecutionException {
-        runAdbShell("am", "start", "-n", packageName);
+    public void startPackage(String packageName) throws InterruptedException, AndroidDebugBridgeDeviceException,
+            AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
+        Boolean fallbackPackage = fallbacks.getOrDefault(HDMI_STATE_CHANNEL, false);
+        String result = runAdbShell("am", "start", "-n", packageName);
+        if (fallbackPackage == false && result.contains("Error")) {
+            fallbacks.put(START_PACKAGE_CHANNEL, true);
+            String fallback = runAdbShell("monkey", "--pct-syskeys", "0", "-p", packageName, "-v", "1");
+            if (fallback.contains("monkey aborted")) {
+                throw new AndroidDebugBridgeDeviceException("Unable to open package: " + packageName);
+            }
+        }
+        throw new AndroidDebugBridgeDeviceReadException(START_PACKAGE_CHANNEL, result);
     }
 
     public void stopPackage(String packageName)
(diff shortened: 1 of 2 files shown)


bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -117,9 +121,18 @@ public void sendText(String text)
         runAdbShell("input", "text", URLEncoder.encode(text, StandardCharsets.UTF_8));
     }
```
GiviMAD, line 123: I have addressed this in to a separate pr, so you don't need to worry about it. I've also added a regex for the packageName to avoid passing weird strings to the command.

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -170,12 +183,14 @@ public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, Inte
 
     public Optional<Boolean> isHDMIOn() throws InterruptedException, AndroidDebugBridgeDeviceException,
             AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
+        Boolean fallbackHDMI = fallbacks.getOrDefault(HDMI_STATE_CHANNEL, false);
         String result = runAdbShell("cat", "/sys/devices/virtual/switch/hdmi/state");
```
GiviMAD, line 187: You are not avoiding running the command once the fallback is enabled. I'll move the logcat command into a separate private method and call it directly once fallback enabled becomes true.

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -170,12 +183,14 @@ public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, Inte
...
-        if (result.equals("0") || result.equals("1")) {
+        if (fallbackHDMI == false && result.equals("0") || result.equals("1")) {
             return Optional.of(result.equals("1"));
-        } else if (result.isEmpty()) {
+        } else if (fallbackHDMI == false && result.isEmpty()) {
             return Optional.empty();
         } else {
```
GiviMAD, line 192: I think that we should look for a concrete output in the command to enable the fallback, something like an error output that said file is not there.

Trinitus01: The result empty check is not for the fallback, it is to keep the state if nothing returns. If something returns it must be a valid info or an error. If it is empty the log hasnt the value anymore and it waits for a new state. So it is really necessary to check for empty.

GiviMAD: That's right, my bad. 
We can also have thrown a AndroidDebugBridgeDeviceReadException as for other inconclusive results. We can let it as is if you prefer but I think it will be more accurate because in this way the command can fail forever and the user is not informed. What do you think?

Trinitus01: Maybe i have to describe a little bit more. Lets say the logcat show up max 30 Lines in a loop. Now the last lines shows the status "hdmi on". After this the hdmi stays on for lets say 30 minutes. But after 10 minutes the logcat doesnt show the status "hdmi on" anymore because of the log rotation. Why to inform the user every 30 seconds for the timespan of 20 minutes that the binding has no new state. It is not an error or something. After 30 minutes the user turn off the tv and the logcat shows hmdi off, now the state is off, and so on. For me there is no point the state becomes out of sync or stays in a wrong state. Maybe i missunderstood your request? Maybe because i transfered this behavior also to the other commands without logcat or log rotation? This isnt necessary like it is for the logcat but i thought if a command on a shell will not return any output it is also not an error because else it would return something. So on a closer look this is only necessary for the result of the logcat commands.

Trinitus01: But i slowly understand whats your point. For the logcat part its a big problem to know if there is no status at the moment or if there never will be a status and nobody knows. Maybe we can inform the user there is no new status but not that often or not on any refresh? Still thinking about a solution.

GiviMAD: I'm talking about the return of the cat command you use at first not logcat command. What you said applies to the output of the logcat command but I don't think it's the same for the cat output.
Anyway I can't test that on my device:
```
mantis:/ $ cat /sys/devices/virtual/switch/hdmi/state
/system/bin/sh: cat: /sys/devices/virtual/switch/hdmi/state: Permission denied
```

GiviMAD: I think you could check for "No such file or directory" or "Permission denied" in the output in order to enable the fallback.

J-N-K: If you only need the value as a state, why not directly return `State` instead of the `Optional<Boolean>` and ude `UndefType.NULL` if there is no value. You could check for that before posting the update.  You could also make a difference with `UndefType.UNDEF` if you really don't know.

Trinitus01: > I'm talking about the return of the cat command you use at first not logcat command. What you said applies to the output of the logcat command but I don't think it's the same for the cat output.
> Anyway I can't test that on my device:
> 
> ```
> mantis:/ $ cat /sys/devices/virtual/switch/hdmi/state
> /system/bin/sh: cat: /sys/devices/virtual/switch/hdmi/state: Permission denied
> ```

Yes you are right. The empty check is not necessary for other checks than logcat. But, i also think it is not necessary to check for the string "permission denied" or other results because if the result does not return 0 or 1 after the split and if the fallback won't returns "SWITCH_STATE=" it should throw a AndroidDebugBridgeDeviceReadException. So lets check for the common returns and if not we throw the exception?

Trinitus01: For example: If /sys/devices/virtual/switch/hdmi/state does not return any of the expected results, throw an exception because it seems the device does not support the command.

Trinitus01: removed all the fallback remember suff

Trinitus01:
@J-N-K Please merge the PR because now it only contains a missing refresh of the hdmi channel. We will continue on @GiviMAD PR for the "remember fallback" stuff.

J-N-K (approved):

PR #49, opened 2021-03-26
[androiddebugbridge] add fallback to start package channel

GiviMAD:
Add fallback to the start package channel, using monkey command to open applications instead of am on devices which doesn't support the -n parameter (devices under android 8).

### DIFF
--- a/bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
+++ b/bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
@@ -19,6 +19,7 @@ import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBri
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.MEDIA_CONTROL_CHANNEL;
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.MEDIA_VOLUME_CHANNEL;
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.SCREEN_STATE_CHANNEL;
+import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.START_PACKAGE_CHANNEL;
 import static org.smarthomej.binding.androiddebugbridge.internal.AndroidDebugBridgeBindingConstants.WAKE_LOCK_CHANNEL;
 
 import java.io.*;
@@ -31,6 +32,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Arrays;
 import java.util.Base64;
+import java.util.HashMap;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.*;
@@ -86,7 +88,7 @@ public class AndroidDebugBridgeDevice {
     private String ip = "127.0.0.1";
     private int port = 5555;
     private int timeoutSec = 5;
-    private boolean fallbackStartPackage = false;
+    private HashMap<String, FallbackModes> channelFallbackMap = new HashMap<>();
     private @Nullable Socket socket;
     private @Nullable AdbConnection connection;
     private @Nullable Future<String> commandFuture;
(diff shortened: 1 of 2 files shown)


Trinitus01:
Does the openURL work on your device? (Because of the -a and -d parameters? Or does it also need a fallback?

Trinitus01:
Should i remove all the result empty checks (except the one for the logcat result) in a new pr or will you change it?

GiviMAD:
Hi,
The openURL channel is working correctly on my device (android 7.1.2) so no need for a fallback on my side.
And yes I think that we can remove those empty results and raised exception exceeding the ones that depends on a log rotation as you explained to me earlier on a separate pr.

Trinitus01 (changes requested):

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -84,6 +86,7 @@
     private String ip = "127.0.0.1";
     private int port = 5555;
     private int timeoutSec = 5;
+    private boolean fallbackStartPackage = false;
```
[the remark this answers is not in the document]
J-N-K: If you do that, please name it "allowlist/denylist" or "accesslist/blocklist" instead of "whitelist/blacklist". Thanks.

Trinitus01 (dismissed):

J-N-K:
I‘m not sure why this happened, but can you please rebase on main? There are some changes which do not belong here.

Trinitus01:
There is some robot binding stuff?!?

GiviMAD:
My fault, it's fixed now.

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -167,23 +193,32 @@ public boolean isScreenOn() throws InterruptedException, AndroidDebugBridgeDevic
 
     public Optional<Boolean> isHDMIOn() throws InterruptedException, AndroidDebugBridgeDeviceException,
             AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
+        if (channelFallbackMap.get(HDMI_STATE_CHANNEL) == FallbackModes.LOGCAT) {
```
Trinitus01, line 196: I prefer equals instead of '=='.
But as i read it seems to be fine to use '==' for enum.

Trinitus01 (approved):

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -84,6 +88,7 @@
     private String ip = "127.0.0.1";
     private int port = 5555;
     private int timeoutSec = 5;
+    private HashMap<String, FallbackModes> channelFallbackMap = new HashMap<>();
```
J-N-K, line 91:
```suggestion
    private Map<String, FallbackModes> channelFallbackMap = new HashMap<>();
```

J-N-K (changes requested):

Trinitus01:
For isHDMIOn() it seems the first command can fail sometimes but the fallback command is not supported by the device. If this happens it never goes back to the first command. I don't know why it can fail, maybe the result is empty between it switches from 0 to 1 and the cat fetches the result between the two states?!? To get out of the fallback we could remove the fallback from the map if the fallback method runs into an exception. So it will try both (or more fallbacks in the future) again. What do you think about that? How to ensure it tries the first command again if both method fail?

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -167,23 +193,32 @@ public boolean isScreenOn() throws InterruptedException, AndroidDebugBridgeDevic
...
+        String result = runAdbShell("logcat", "-d", "|", "grep", "hdmi", "|", "grep", "SWITCH_STATE=", "|", "tail",
+                "-1");
+        if (result.contains("SWITCH_STATE=")) {
+            return Optional.of(result.contains("SWITCH_STATE=1"));
+        } else if (result.isEmpty()) {
+            // IF THE DEVICE DO NOT SUPPORT THIS VALUE IN LOGCAT THE USER WILL NEVER KNOW THE CHANNEL WON'T WORK
+            // FIND A BETTER SOLUTION
+            return Optional.empty();
+        }
+        throw new AndroidDebugBridgeDeviceReadException(HDMI_STATE_CHANNEL, result);
```
Trinitus01, line 219:
```suggestion
        channelFallbackMap.remove(HDMI_STATE_CHANNEL);
        throw new AndroidDebugBridgeDeviceReadException(HDMI_STATE_CHANNEL, result);
```

Trinitus01: To the case it never tries the first method again and the fallback is not supported by the device.

Trinitus01 (changes requested):

Trinitus01 (approved):

J-N-K (approved):

PR #53, opened 2021-03-28
[irobot] fix MQTT QoS on newer openHAB cores

J-N-K:
In older (<3.1) versions of openHAB core the wrong QoS level was sent (0 instead of 1 and vice-versa). After this has been fixed the QoS send in the binding needs to be adapted. 


J-N-K:
@falkena:  FYI: This should make it work for 3.0 and 3.1

PR #70, opened 2021-04-17
[androiddebugbridge] shutdown channel, naming and improvements

Trinitus01:
ported: shutdown channel
ported: tap event pattern
ported: some variable names
changed: mouse-tap channel to tap channel
changed: package name pattern for extended package names

ported from: https://github.com/openhab/openhab-addons/pull/10497

Also-By: Miguel <miguelwork92@gmail.com>


J-N-K (approved):
Thanks. I'll wait until the original PR is merged so we can port changes that may occur in the review process and thus keep the user interface consistent.

PR #88, opened 2021-05-01
[notificationsforfiretv] thing status depending on the last request

Trinitus01:
added: set thing status depending on the last request

Inspired by your last commit to the http binding.
What do you think about setting the thing status depending on the last request?
It also do a request without a real notification on binding initialisation.

### DIFF
--- a/bundles/org.smarthomej.binding.notificationsforfiretv/src/main/java/org/smarthomej/binding/notificationsforfiretv/internal/NotificationsForFireTVHandler.java
+++ b/bundles/org.smarthomej.binding.notificationsforfiretv/src/main/java/org/smarthomej/binding/notificationsforfiretv/internal/NotificationsForFireTVHandler.java
@@ -38,6 +38,7 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.thing.Thing;
 import org.openhab.core.thing.ThingStatus;
+import org.openhab.core.thing.ThingStatusDetail;
 import org.openhab.core.thing.binding.BaseThingHandler;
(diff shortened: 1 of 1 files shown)


bundles/org.smarthomej.binding.notificationsforfiretv/src/main/java/org/smarthomej/binding/notificationsforfiretv/internal/NotificationsForFireTVHandler.java
```diff
@@ -91,29 +93,40 @@ public boolean sendNotification(@Nullable String msg, @Nullable String filename,
...
             notificationsForFireTVConnection.finish();
 
+            // UPDATE STATUS
+            updateStatus(ThingStatus.ONLINE);
+
             return true;
-        } catch (Exception e) {
-            logger.warn("Unable to connect: {}", e.getMessage());
+        } catch (IOException e) {
+            logger.debug("Unable to send notification: {}", e.getMessage());
```
J-N-K, line 123:
```suggestion
            logger.debug("Unable to send notification: ", e);
```

Either log the full stack trace here (if it add useful information) or remove the logging. The change of the thing status is logged anyway, so updating the thing status already logs `e.getMessage()`

J-N-K (changes requested):

bundles/org.smarthomej.binding.notificationsforfiretv/src/main/java/org/smarthomej/binding/notificationsforfiretv/internal/NotificationsForFireTVHandler.java
Trinitus01: done

J-N-K (approved):

PR #95, opened 2021-05-09
[transform.basicprofiles] Initial contribution of Basic Profiles bundle

cweitkamp:
- Initial contribution of Basic Profiles bundle


J-N-K (changes requested):

bundles/org.smarthomej.binding.basicprofiles/src/main/feature/feature.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<features name="org.smarthomej.binding.basicprofiles-${project.version}" xmlns="http://karaf.apache.org/xmlns/features/v1.4.0">
+	<repository>mvn:org.openhab.core.features.karaf/org.openhab.core.features.karaf.openhab-core/${ohc.version}/xml/features</repository>
+
+	<feature name="openhab-binding-basicprofiles" description="Basic Profiles Binding" version="${project.version}">
```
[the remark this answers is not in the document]
cweitkamp: If I have to choose between "binding" and "transformation" the latter would fit better.

pom.xml
```diff
@@ -348,6 +348,7 @@ Import-Package: \\
           <version>2.10.3</version>
           <configuration>
             <failOnError>!${quality.skip}</failOnError>
+            <javadocExecutable>${java.home}/bin/javadoc</javadocExecutable>
```
cweitkamp, line 351: Should I revert this?

bundles/org.smarthomej.transform.basicprofiles/README.md
```diff
@@ -0,0 +1,56 @@
+# BasicProfiles Binding
```
cweitkamp, line 1: I am not  good writer ...

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/BasicProfilesConstants.java
```diff
@@ -0,0 +1,26 @@
...
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.smarthomej.transform.basicprofiles.internal;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * The {@link BasicProfilesConstants} class defines common constants, which are used across the whole binding.
```
cweitkamp, line 18:
```suggestion
 * The {@link BasicProfilesConstants} class defines common constants, which are used across the whole bundle.
```

bundles/org.smarthomej.transform.basicprofiles/src/main/resources/OH-INF/binding/binding.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="basicprofiles" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:binding="https://openhab.org/schemas/binding/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/binding/v1.0.0 https://openhab.org/schemas/binding-1.0.0.xsd">
+
+	<name>Basic Profiles</name>
+	<description>Binding for basic profiles which doesn't fit into a specific other binding.</description>
```
cweitkamp, line 7:
```suggestion
	<description>Bundle for basic profiles which doesn't fit into a specific bundle.</description>
```

🤔 Hm, is this needed at all for transformations?

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/RoundStateProfile.java
```diff
@@ -0,0 +1,136 @@
...
+        // }
+        // } else
+        Integer configScale = config.scale;
+        if (configScale instanceof Number) {
+            localScale = ((Number) configScale).intValue();
+        } else {
+            logger.error("Parameter 'scale' is not of type String or Number.");
+        }
+
+        if (localScale < 0) {
```
J-N-K, line 70: Why is this restricted to non-negative numbers? I always thought that a negative scale is allowed and e.g. a scale of -2 round 1234 to 1200. This could also be useful.

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/RoundStateProfile.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Override
+    public ProfileTypeUID getProfileTypeUID() {
+        return ROUND_UID;
+    }
+
+    @Override
+    public void onStateUpdateFromItem(State state) {
+        callback.sendUpdate((State) applyRound(state));
```
J-N-K, line 97: Is that correct? I thought that `sendUpdate` updates the item's state. But shouldn't this result in a state that is processed by the handler?

bundles/org.smarthomej.transform.basicprofiles/src/main/resources/OH-INF/binding/binding.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
J-N-K, line 1: Other transformations don't have this file. It can be removed.

bundles/org.smarthomej.transform.basicprofiles/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,116 @@
...
+				<option value="FASTFORWARD">FASTFORWARD</option>
+				<option value="STOP">STOP</option>
+				<option value="MOVE">MOVE</option>
+				<option value="UP">UP</option>
+				<option value="DOWN">DOWN</option>
+			</options>
+		</parameter>
+	</config-description>
+
+	<config-description uri="profile:deconz:repeated-command">
```
J-N-K, line 95: this looks wrong here

bundles/org.smarthomej.transform.basicprofiles/src/main/resources/OH-INF/i18n/basicprofiles.properties
```diff
@@ -0,0 +1,43 @@
+profile-type.basic-profiles.battery-low.label = Battery Low
```
J-N-K, line 1: Please check these files. It looks like the languages are mixed (e.g. "Bildet bei Änderung... " in this file or  "Scale (non-negative number) to be applied on the state." in the german file).

bundles/org.smarthomej.transform.basicprofiles/src/test/java/org/smarthomej/transform/basicprofiles/internal/factory/BasicProfilesFactoryTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+        Collection<ProfileType> supportedProfileTypes = profileFactory.getProfileTypes(null);
+        assertThat(supportedProfileTypeUIDs.size(), is(supportedProfileTypes.size()));
+
+        for (ProfileType profileType : supportedProfileTypes) {
+            assertThat(supportedProfileTypeUIDs.contains(profileType.getUID()), is(true));
+        }
+    }
+
+    // @Test
```
J-N-K, line 78: remove unused code

J-N-K (changes requested):
Thanks. These look like great additions. I have left some comments.

bundles/org.smarthomej.transform.basicprofiles/README.md
```diff
@@ -0,0 +1,62 @@
+# Basic Profiles
+
+This bundle provides a list of useful Profiles.
+
+## Battery Low Profile
```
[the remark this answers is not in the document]
cweitkamp: Yes, of course. I think this will end I the same discussion I had with Kai about this. The result was the Hysteresis Profile enhanced with an `inverted` configuration parameter. Long story short: In general this Profile is just a shortcut for another existing Profile.

bundles/org.smarthomej.transform.basicprofiles/README.md
```diff
@@ -0,0 +1,62 @@
...
+| Configuration Parameter | Description | Type |
+|-|
+| `scale` | Scale (non-negative number) to indicate the resulting number of decimal places (min: 0, max: 16) **mandatory**. | integer |
+| `mode` | Rounding mode to be used (e.g. "UP", "DOWN", "CEILING", "FLOOR", "HALF_UP" or "HALF_DOWN" (default: "HALF_UP", min: 0, max: 16). | text |
+
+### Full Example
+
+```java
+Number roundedNumber { channel="xxx" [profile="basic-profiles:round", scale=1] }
+```
```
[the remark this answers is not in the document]
cweitkamp: I will crawl through the code and remove everything which is not working. Probably only Round and Invert will remain. Timestamp Offset is in the OHC queue. I hope it will be accepted over there.

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/RoundStateProfile.java
cweitkamp: No. Round is a one-way Profile. It should be applied only in updates / commands from handler towards Item.

bundles/org.smarthomej.transform.basicprofiles/README.md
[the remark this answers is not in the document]
J-N-K: I really like it, because it‘s so simple and easy to understand. Together with the invert profile and the chain profile (see my PR) you can use it in any way. No need for more configuration. It‘s just the name.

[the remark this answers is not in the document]
cweitkamp: :+1: Makes sense. I will change it.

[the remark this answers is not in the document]
cweitkamp: > Together with the invert profile and the chain profile (see my PR) you can use it in any way.

This is not clear to me. You Chain Profile only supports chaining Transformations. Did I miss something?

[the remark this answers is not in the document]
J-N-K: You are right. I was think of chaining profiles but that is a different issue from chaining transformations. Interesting idea though.

cweitkamp:
One more question. Should I contribute against the "main" branch?

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/RoundStateProfile.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private final Logger logger = LoggerFactory.getLogger(RoundStateProfile.class);
+
+    public static final String PARAM_SCALE = "scale";
+    public static final String PARAM_MODE = "mode";
+
+    private final ProfileCallback callback;
+
+    final int scale;
+    final RoundingMode roundingMode;
```
J-N-K, line 51:
```suggestion
    private final int scale;
    private final RoundingMode roundingMode;
```

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/RoundStateProfile.java
J-N-K: Can you comment here?

J-N-K (changes requested):
Just two comments. Otherwise looks good. Jenkins fails unrelated. 

Regarding the branch: I sync both branches manually by cherry-pciking the commits, so it doesn't matter wher you contribute. Just use what better fits your need.

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/RoundStateProfile.java
cweitkamp: I am using package visibility here because the related unit test checks for correct parsing of the configuration parameters.

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/RoundStateProfile.java
cweitkamp: Sry. I missed this comment. Might be useful too. Will change it.

cweitkamp: Done.

J-N-K (approved):

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/RoundStateProfile.java
J-N-K: @cweitkamp I would like to come back to this comment, as the current implementation causes issues with latest core changes. I also came across this when using the invert/negate profile.

As a result of this line we end up in an endless loop: 

1. The item updates its state to `123.12` and sends an `ItemStateUpdatedEvent` with a state of `123.12`.
2. The `CommunicationManager` forwards this state update to the and according to its configuration the profile rounds that to `123` and forwards that to the callback's `sendUpdate` method.
3. The callback issues an `ItemStateEvent` which is received by the item, which sets its state to `123`. After that, it sends an `ItemStateUpdatedEvent` with a state of `123`.
4. The `CommunicationManager` receives that, ... (start again at 2).

the same issue arises with the invert profile, because `CLOSED` is inverted to `OPEN` and after updating the item state to `OPEN` it is again inverted to `CLOSED`, resulting in the same update loop.

We can't prevent that by checking event `source`, because the `ItemUpdater` (that receives the `ItemStateEvent`) calls `GenericItem.setState` (which creates the `ItemStateUpdatedEvent/ItemStateChangedEvent`) and the `source` is lost (otherwise we could have set the `source` of the `ItemStateUpdatedEvent` to the same value as the `ItemStateEvent`.

IMO a profile should only change values that are transferred from the handler to the item (or from the item to the handler), not interact with the item state itself.

LukasA83: And, obviously, it is right now a blocking issue to migrate to OH 4 with M2 or newer snapshots, so maybe this can be resolved somehow.

cweitkamp: Your explanation sounds reasonable. I won't argue against it. Feel free to change/remove this part of code.

cweitkamp: After upgrading to latest OH milestone I am suffering from this myself. I submitted https://github.com/smarthomej/addons/pull/478 to fix this.

PR #102, opened 2021-05-14
[tr064] Added unit test for CallListType

cweitkamp:
- Added unit test for CallListType


bundles/org.smarthomej.binding.tr064/src/test/java/org/smarthomej/binding/tr064/internal/soap/CallListTypeTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+        public ParameterSet(String dateString, String resultingDateString) {
+            this.dateString = dateString;
+            this.resultingDateString = resultingDateString;
+        }
+    }
+
+    public static Collection<Object[]> parameters() {
+        return List.of(new Object[][] { //
+                { new ParameterSet("15.02.21 08:20", "2021-02-15T08:20:00+01") }, //
```
J-N-K, line 52: Assuming `+01` as timezone looks wrong. In fact the code assume sthat the FritzBox TZ is the same as the local systems TZ. I guess that is why it fails.

PR #104, opened 2021-05-15
fix spotless issues

J-N-K:
- add {} for if (http)
- remove unnecessary workaround in knx (already fixed in calimero 2.4)
- remove unused logger (notificationsforfiretv)
- fix local variable naming (onewire)
- fix markdown in documentation (telenot)


Trinitus01 (commented):
Looks fine. Seems i forgot to remove the logger which is not needed anymore.

PR #105, opened 2021-05-16
[transform.math] add divide

J-N-K:
This is useful for e.g. to display the remaining time of the dishwasher in minutes instead of seconds. This could also be achieved by multiplying with `0.1666667` but having a dedicated transformation for that is more convenient.

### DIFF
--- a/bundles/org.smarthomej.transform.math/README.md
+++ b/bundles/org.smarthomej.transform.math/README.md
@@ -9,6 +9,7 @@ Example Items:
 ```
 Number multiply "Value multiplied by [MULTIPLY(1000):%s]" { channel="xxx" }
 Number add "Value added [ADD(5.1):%s]" { channel="xxx" }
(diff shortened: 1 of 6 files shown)


cweitkamp:
Two questions: Should we take care of division by 0? Or handle non-terminating decimal expansion like 1/3? Do we need a unit test for both?

J-N-K:
Good point. BigDecimal throws an ArithmeticException for a division by zero, so I decided to catch that in the AbstractMathTransformationService. This also takes care of other errors that could occur in later additions.

Regarding the non terminating expansion: This is a bit difficult, because nobody knows the needed precision. It could be that the transformed value is fed into a Number item and used for other calculations and cutting decimals results in loss of needed precision.

cweitkamp (approved):
IIRC the `devide()` method will throw an `ArithmeticException` too if there are non terminating decimal issue. Thus we should be in safe haven for now.

PR #110, opened 2021-05-22
[tr064] improve channel parameter processing

### LINKED ISSUE #109: [tr064] Warning for invalid configuration parameter if empty
I often see the following warning in my log files. In general the message is okay because I did not add a value for the checked parameter.

```
2021-05-22 14:28:24.112 [WARN ] [hab.binding.tr064.internal.util.Util] - Removing while processing callList, does not match pattern [0-9]+, check config.
```

On first setup of the Thing I set a value of 7 days for this configuration parameter via OH UI. After some time I did not need this information anymore and tried to remove the configuration and the related Channel by clearing the configuration parameter by clicking on the grey cross in the configuration page of the UI. This results in an empty parameter on the Thing in my JSON storage. Now the mentioned validation fails. Wdyt? Can we improve this?

Additionally we might want to improve the message itself.

J-N-K:
Fixes #109 


cweitkamp (commented):
Thanks. One minor comment.

cweitkamp (approved):

PR #112, opened 2021-05-24
[androiddebugbridge] double check result before update state 

Trinitus01:
added: double check result before update state

I recognised that some devices have some state fluctuations and the state turns to on although it is still off.
It seems Miguel already recognised this behavior and added another workaround to the binding in the openHAB project.
I prefer to do the double check of the result in the "handleCommandInternal"-method instead of the "refreshStatus"-method.

Signed-off-by: Tom Blum trinitus01@googlemail.com

### DIFF
--- a/bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
+++ b/bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
@@ -19,7 +19,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -197,14 +196,13 @@ public class AndroidDebugBridgeHandler extends BaseThingHandler {
                 break;
             case HDMI_STATE_CHANNEL:
                 if (command instanceof RefreshType) {
-                    Optional<Boolean> hdmiState = adbConnection.isHDMIOn();
-                    boolean lastHDMIState = (boolean) channelLastStateMap.getOrDefault(HDMI_STATE_CHANNEL, false);
-                    if (hdmiState.isPresent()) {
-                        if (hdmiState.get().equals(lastHDMIState)) {
-                            updateState(channelUID, OnOffType.from(hdmiState.get()));
+                    adbConnection.isHDMIOn().ifPresent(hdmiState -> {
+                        boolean lastHDMIState = (boolean) channelLastStateMap.getOrDefault(HDMI_STATE_CHANNEL, false);
+                        if (hdmiState.equals(lastHDMIState)) {
+                            updateState(channelUID, OnOffType.from(hdmiState));
                         }
-                        channelLastStateMap.put(HDMI_STATE_CHANNEL, hdmiState.get());
-                    }
+                        channelLastStateMap.put(HDMI_STATE_CHANNEL, hdmiState);
+                    });
                 }
                 break;
         }


bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
```diff
@@ -163,22 +167,44 @@ private void handleCommandInternal(ChannelUID channelUID, Command command)
...
                 if (command instanceof RefreshType) {
-                    adbConnection.isHDMIOn().ifPresent(state -> updateState(channelUID, OnOffType.from(state)));
+                    Optional<Boolean> hdmiState = adbConnection.isHDMIOn();
+                    boolean lastHDMIState = (boolean) channelLastStateMap.getOrDefault(HDMI_STATE_CHANNEL, false);
+                    if (hdmiState.isPresent()) {
+                        if (hdmiState.get().equals(lastHDMIState)) {
+                            updateState(channelUID, OnOffType.from(hdmiState.get()));
+                        }
+                        channelLastStateMap.put(HDMI_STATE_CHANNEL, hdmiState.get());
+                    }
```
J-N-K, line 207:
```suggestion
                    adbConnection.isHDMIOn().ifPresent(hdmiState -> {
                        boolean lastHDMIState = (boolean) channelLastStateMap.getOrDefault(HDMI_STATE_CHANNEL, false);
                        if (hdmiState.get().equals(lastHDMIState)) {
                            updateState(channelUID, OnOffType.from(hdmiState.get()));
                        }
                        channelLastStateMap.put(HDMI_STATE_CHANNEL, hdmiState.get());
                    });
```

J-N-K (changes requested):
Otherwise looks good.

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
Trinitus01: done

cweitkamp:
> It seems Miguel already recognised this behavior and added another workaround to the binding in the openHAB project.

The binding over here seem to be in a much better state than in openHAB Add-ons repository. Are there any features missing here which are part of OH version?

Trinitus01:
> > It seems Miguel already recognised this behavior and added another workaround to the binding in the openHAB project.
> 
> The binding over here seem to be in a much better state than in openHAB Add-ons repository. Are there any features missing here which are part of OH version?

No, there is no feature missing in this binding. But there are features missing in the OH version, like the fallback for some devices.

J-N-K:
Manually verified sign-off

J-N-K (approved):

bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
```diff
@@ -163,22 +166,43 @@ private void handleCommandInternal(ChannelUID channelUID, Command command)
                 }
             case WAKE_LOCK_CHANNEL:
                 if (command instanceof RefreshType) {
-                    updateState(channelUID, new DecimalType(adbConnection.getPowerWakeLock()));
+                    int wakeLockState = adbConnection.getPowerWakeLock();
+                    int lastWakeLockState = (int) channelLastStateMap.getOrDefault(WAKE_LOCK_CHANNEL, 0);
+                    if (wakeLockState == lastWakeLockState) {
+                        updateState(channelUID, new DecimalType(wakeLockState));
+                    }
+                    channelLastStateMap.put(WAKE_LOCK_CHANNEL, wakeLockState);
```
cweitkamp, line 174: I have to admit, that I do not like these changes. Especially for the wakeLock Channel. I tested it and it really worsened my user experience.

Additionally in my POV debouncing has not to be placed into a Binding. Therefore I created https://github.com/smarthomej/addons/pull/113. A long overdue feature for openHAB.

Please revert these changes. One thing which can be reused is the internal State cache. e.g. the wakeLock is retrieved twice if a user linked the media-control Channel and configured a wakeLock-based "PLAY" detection. We can reorder the refresh requests in [AndroidDebugBridgeHandler# refreshStatus()](https://github.com/smarthomej/addons/blob/abc6a63ed7e4bd83d8ecdb1b5674639d25e59ebe/bundles/org.smarthomej.binding.androiddebugbridge/src/main/java/org/smarthomej/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java#L369) to request the wakeLock first, store / cache it internally and reuse it for mediaControl detection.

One could add an `ExpiringCacheMap` for adb calls too.

Trinitus01: Hi,

for me the double check is only necessary for the hdmi channel. As i know for Miguel it is only necessary for the screen channel. I transfered the double check to all state channels, so maybe this was a bad decision. I will revert it fot the wake lock tomorrow.

br

cweitkamp: Thanks for looking into it. What do you think about https://github.com/smarthomej/addons/commit/080e29e37e8dee4f39ad4a7e74cb6f62477f3d9b?

Trinitus01: For me it looks fine, no doubts about this strategy from my side. Thanks for the improvements, too.

PR #124, opened 2021-06-07
[amazonechocontrol] fix cycling of UNKOWN/ONLINE for smarthome devices

J-N-K:
The code did not properly handle the case where the capability is present but not the individual device and thus cycled between UNKNOWNand ONLINE.

### DIFF
--- a/bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/handler/SmartHomeDeviceHandler.java
+++ b/bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/handler/SmartHomeDeviceHandler.java
@@ -18,7 +18,6 @@ import static org.smarthomej.binding.amazonechocontrol.internal.smarthome.Consta
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -63,7 +62,6 @@ import org.smarthomej.binding.amazonechocontrol.internal.smarthome.HandlerBase.U
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
-import com.google.gson.JsonNull;
 import com.google.gson.JsonObject;
 
 /**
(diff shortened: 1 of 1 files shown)


bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/handler/SmartHomeDeviceHandler.java
```diff
@@ -185,62 +179,65 @@ public void updateChannelStates(List<SmartHomeBaseDevice> allDevices,
...
         for (HandlerBase handlerBase : handlers.values()) {
             UpdateChannelResult result = new UpdateChannelResult();
             for (String interfaceName : handlerBase.getSupportedInterface()) {
                 List<JsonObject> stateList = mapInterfaceToStates.get(interfaceName);
                 if (stateList != null) {
                     try {
                         handlerBase.updateChannels(interfaceName, stateList, result);
-                    } catch (Exception e) {
+                    } catch (RuntimeException e) {
                         // We catch all exceptions, otherwise all other things are not updated!
```
mgeramb, line 241: Not sure if it is a good idea to change to RuntimeException here. If anything else happens which is not expected, all other things will not be updated. This was the reason for catching Exception.

bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/handler/SmartHomeDeviceHandler.java
```diff
@@ -255,11 +252,7 @@ public void updateChannelStates(List<SmartHomeBaseDevice> allDevices,
             }
         }
 
-        if (stateFound) {
-            updateStatus(ThingStatus.ONLINE);
```
mgeramb, line 259: maybe we a debug trace instead of the updateStatus would be better. Removing this "if" complete would make diagnostic much harder.

mgeramb (changes requested):

bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/handler/SmartHomeDeviceHandler.java
J-N-K: The logging in l. 228 is exactly that. It just exists early and skips the processing the lookup of applianceIds and interfaces if the state array is empty.

bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/handler/SmartHomeDeviceHandler.java
J-N-K: But all exceptions that do not inherit from RuntimeException are checked exceptions and thus the compiler makes sure that they are catched.

mgeramb: Ok, I see, this feature does not exist in C#, so I'am not familiar with it.

PR #126, opened 2021-06-07
[tr064] allow skipping digits from left in phonebook lookup

J-N-K:
Positive `matchCount` values match a defined number of digits from the right (end) of the phonenumber. Negative `matchCount` values skip digits from the left (e.g. for removing international or national prefixes).


J-N-K:
Functionality is also community-approved.

cweitkamp (approved):
Code looks good.

If two matching phone numbers will be found the first one will be returned. Is that correct?

J-N-K:
True. "first" is however not easy to determine, since in the case of multiple phonebooks the ordering depends on the internal ordering of the phonebooks in the FritzBox.

PR #131, opened 2021-06-12
[mail] add mail content processing

J-N-K:
A new channel-type `content` allows the processing of mail contents. Mails can be filtered by subject and transformed before the item-state is updated.

This has been request in the German Facebook user group.


cweitkamp (dismissed):
LGTM. Nothing critical to complain about.

PR #138, opened 2021-06-21
[deconz] Added support for publishing ChannelDescriptionChangedEvents

cweitkamp:
- Added support for publishing `ChannelDescriptionChangedEvents`

Related to openhab/openhab-core#1505
Depends on https://github.com/openhab/openhab-core/pull/2410


bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/handler/GroupThingHandler.java
```diff
@@ -241,10 +240,8 @@ public void messageReceived(DeconzBaseMessage message) {
             Map<String, String> scenes = groupMessage.scenes.stream()
                     .collect(Collectors.toMap(scene -> scene.id, scene -> scene.name));
             ChannelUID channelUID = new ChannelUID(thing.getUID(), CHANNEL_SCENE);
-            commandDescriptionProvider.setDescription(channelUID,
-                    CommandDescriptionBuilder.create().withCommandOptions(groupMessage.scenes.stream()
-                            .map(scene -> new CommandOption(scene.name, scene.name)).collect(Collectors.toList()))
```
cweitkamp, line 246: Was this mapping `new CommandOption(scene.name, scene.name)` by intention? Or was it a mistake?

J-N-K: Since the scene id is an arbitrary number which is hard to remember, I thought it would be a good idea to have the name as command (and label).  If this is changed, the GroupThingHandler needs to be adjusted, too (it expects the scene name as command).

It's debatable if using the id is better, but the deconz API ensures that the name is also unique (storing a new scene with the name of an existing name will overwrite the old scene and re-use the id).

bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/Util.java
```diff
@@ -106,7 +105,7 @@ public static DateTimeType convertTimestampToDateTime(String timestamp) {
      * @param value the value to find in the map
      * @return Stream of all keys for the value
      */
-    public static <@NonNull K, @NonNull V> Stream<K> getKeysFromValue(Map<K, V> map, V value) {
+    public static <K, V> Stream<K> getKeysFromValue(Map<K, V> map, V value) {
```
J-N-K, line 108: Is there a reason why you changed that? We have to decide here: either a SAT warning (if we use @NonNull) or a compile warning (if we don‘t use it).

bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/handler/GroupThingHandler.java
cweitkamp: Thanks for clarification. I should have checked the `handleCommand()` code to see how it is working.

bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/Util.java
cweitkamp: Reverted. I removed it because of SAT warning.

J-N-K:
Can you rebase on 3.2.x? We can't merge this to main because the new methods and fields are not available in OHC 3.0.x. I tried just changing the base branch but that resulted in a very big changeset.

J-N-K (approved):

cweitkamp:
> Can you rebase on 3.2.x?

Done.

J-N-K:
bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/DeconzDynamicStateDescriptionProvider.java:[77,55] Potential null pointer access: this expression has a '@Nullable' type

@cweitkamp, can you fix that?

cweitkamp:
> can you fix that?

Done.

J-N-K (approved):

PR #155, opened 2021-07-06
[deconz] Use ChannelBuilder created by ThingHandlerCallback

cweitkamp:
- Use `ChannelBuilder` created by `ThingHandlerCallback`

### DIFF
--- a/bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/handler/SensorThermostatThingHandler.java
+++ b/bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/handler/SensorThermostatThingHandler.java
@@ -34,9 +34,8 @@ import org.openhab.core.library.types.StringType;
 import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.thing.Thing;
 import org.openhab.core.thing.ThingTypeUID;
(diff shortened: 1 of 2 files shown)


bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/handler/SensorThermostatThingHandler.java
```diff
@@ -204,10 +203,9 @@ protected void processStateResponse(DeconzBaseMessage stateResponse) {
 
         SensorMessage sensorMessage = (SensorMessage) stateResponse;
         SensorState sensorState = sensorMessage.state;
-        if (sensorState != null && sensorState.windowopen != null && thing.getChannel(CHANNEL_WINDOWOPEN) == null) {
```
J-N-K, line 207: Why did you remove the check if the channel is already present?

cweitkamp: It is done twice. The `createChannel` method checks for existent Channel too:

https://github.com/smarthomej/addons/blob/e79f1bd994c01959e9cd902c045c4874123df18a/bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/handler/DeconzBaseThingHandler.java#L193-L196

PR #164, opened 2021-07-19
[deconz] add locked channel for thermostats

J-N-K:
Some thermostats provide a keylock to protect the settings. The `locked` channel is added when available and is readable and writable.

### DIFF
--- a/bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/thing/sensor-thing-types.xml
+++ b/bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/thing/sensor-thing-types.xml
@@ -525,6 +525,7 @@
 		<item-type>Switch</item-type>
 		<label>Locked</label>
 		<description>Status of this thermostat's childlock.</description>
+		<category>Lock</category>
 	</channel-type>
 	<channel-type id="heatsetpoint">
 		<item-type>Number:Temperature</item-type>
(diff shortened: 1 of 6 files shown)


bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/thing/sensor-thing-types.xml
```diff
@@ -521,6 +521,11 @@
 		<config-description-ref uri="thing-type:deconz:sensor"/>
 	</thing-type>
 
+	<channel-type id="locked">
+		<item-type>Switch</item-type>
+		<label>Locked</label>
+		<description>Status of this thermostat's childlock.</description>
```
cweitkamp, line 527: Maybe add category "Lock"? And define it as an advanced channel type as it is not available an every device?

```suggestion
		<description>Status of this thermostat's childlock.</description>
		<category>Lock</category>
```

cweitkamp (approved):
Code LGTM. One remake on the modeling.

bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/thing/sensor-thing-types.xml
J-N-K: I also thought about that. But after thinking about it, I decided against it. Isn't the category for a "physical" lock (like a doorlock)? This is more a keylock. I didn't find a clear definition when to use the category.


Regarding advanced: since the channel is only added when it is actually available and there are still less than ten channels on the device, this is not needed on my opinion.

cweitkamp: The category has no real meaning on the technical side. It is just there to define a default icon shown in the UIs.

PR #166, opened 2021-07-24
[deconz] change XY conversion

J-N-K:
This changes the HSB-XY conversion to the way Philips suggests for Hue. It also allows a future extension to use the Gamut of the device.

@ebaauw, is the credit given enough?

### DIFF
--- a/bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/ColorUtil.java
+++ b/bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/ColorUtil.java
@@ -1,0 +1,295 @@
+/**
+ * Copyright (c) 2021 Contributors to the SmartHome/J project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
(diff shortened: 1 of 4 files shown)


ebaauw:
> is the credit given enough?

Sure.  Happy to be of help.

J-N-K:
@cweitkamp Did you have a chance to look into this?

cweitkamp:
I do not have any objections on the implementation. But I am still wondering why we do not add a global solution in OHC. Other bindings like Hue itself or ZigBee may benefit from this new way to convert hue specific colors too.

J-N-K:
If someone wants to port this to core, I'm fine with that. The reason to implement it here is to make these improvements available to all users, no matter if they run the upcoming 3.2 or 3.0/3.1. Changes in OHC are not back-ported.

cweitkamp:
I ported this PR to the openHAB Hue binding (see https://github.com/openhab/openhab-addons/compare/main...cweitkamp:bugfix-11031-hue-xy?expand=1). The color conversion itself is much better. But I am struggling with the setting brightness. The unit tests to check for bijective brightness conversion are failing when applying your suggested calculation:

```java
        int brightness = (int) xy[2] * BRIGHTNESS_MAX;
```

The OHC implementation multiplies the `Y` value with the current / expected brightness but that does not work either:

https://github.com/openhab/openhab-core/blob/95da7021d9ce83a2e92a92aab3d79229ab485d92/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/HSBType.java#L375

```java
                new PercentType(Float.valueOf(tmpY * getBrightness().floatValue()).toString()) };
```

Any ideas? Unfortunately I have no Hue bulb connected to my deCONZ setup.

PR #174, opened 2021-08-29
[androiddebugbridge] start intent channel

Trinitus01:
added: start intent channel

adb shell 'am start "intent:#Intent;scheme=customapp;package=com.comp.pac;end"'

Source: https://stackoverflow.com/questions/21034168/android-start-an-activity-from-command-line-using-intent-uri

### DIFF
--- a/bundles/org.smarthomej.binding.androiddebugbridge/pom.xml
+++ b/bundles/org.smarthomej.binding.androiddebugbridge/pom.xml
@@ -16,6 +16,12 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.smarthomej.addons.bundles</groupId>
+      <artifactId>org.smarthomej.commons</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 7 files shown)


Trinitus01:
Maybe someone can create a regex to verify the intent string.

J-N-K:
Code looks good. Please use the update-feature of SmartHome/J:

- add a dependency to the commons bundle like here: https://github.com/smarthomej/addons/blob/3ea2405ff83d35a2f25c2f2d850189a067d2c674/bundles/org.smarthomej.binding.http/pom.xml#L21-L27
- add a file `android.update` in `main/resources/update` with the content: `1;ADD_CHANNEL;start-intent,String,androiddebugbridge:start-intent-channel`
- add `<properties>
			<property name="thingTypeVersion">1</property>
		</properties>` to the thing-type
- inherit the ThingHandler from `UpdatingBaseThingHandler` instead of `BaseThingHandler`

After that all existing things will be automatically upgraded with the new channel, so the user doesn't have to delete and re-add it.

J-N-K (approved):

PR #176, opened 2021-09-09
[amazonechocontrol] Fix announcment with TTS volume

mgeramb:
Signed-off-by: Michael Geramb <mail@michael-geramb.at>


J-N-K (approved):

J-N-K:
Does it break something for the Echo or is it just "not enough" to fix an issue with the echo?

Trinitus01:
> Does it break something for the Echo or is it just "not enough" to fix an issue with the echo?

It fully breaks ACT on the Echo (1. Gen). With this fix ACT does not work anymore on the Echo (1. Gen).

J-N-K:
Superseded by #180

PR #177, opened 2021-09-09
[amazonechocontrol] Fix announcment with TTS volume

mgeramb:
Signed-off-by: Michael Geramb <mail@michael-geramb.at>


J-N-K:
#176 will be cherry-picked to main after merging, no need for a second PR.

PR #199, opened 2021-10-10
[mail] add actions with optional headers

J-N-K:

### DIFF
--- a/bundles/org.smarthomej.binding.mail/README.md
+++ b/bundles/org.smarthomej.binding.mail/README.md
@@ -2,6 +2,9 @@
 
 The Mail binding provides support for sending mails from rules.
 
+***Deprecation warning:*** The action methods `sendMail(...)` and `sendHtmlMail(...)` for sending mails with attachments are deprecated.
+Use `sendMailWithAttachment/sendMailWithAttachments` (or respective methods for HTML mails, see below).
+
 ## Supported Things
 
 There are three things: `smtp`, `imap` and `pop3` which represents respective servers.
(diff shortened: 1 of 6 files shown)


bundles/org.smarthomej.binding.mail/README.md
```diff
@@ -149,9 +156,9 @@ success = mailActions.sendMail("mail1@example.com, mail2@example.com", "Test sub
 ```
 import java.util.List
 
-val List<String> attachmentUrlList = newArrayList(
+val List<String> attachmentUrlList = new ArrayList(
```
cweitkamp, line 159: Does this work without importing it? Can we use `List.of(...)` method?

bundles/org.smarthomej.binding.mail/src/main/resources/OH-INF/i18n/mail.properties
```diff
@@ -16,3 +19,6 @@ sendHTMLAttachmentMessageActionDescription = Sends a HTML mail with an URL attac
 
 sendHTMLAttachmentsMessageActionLabel = send a HTML mail with several attachments
 sendHTMLAttachmentsMessageActionDescription = Sends a HTML mail with several URL attachments.
+
+sendComplexHtmlMessageActionLabel = send a html mail with several attachments and/or additional headers
```
cweitkamp:
```suggestion
sendComplexHtmlMessageActionLabel = send an HTML mail with several attachments and/or additional headers
```

In general we should fix the spelling in this file `a HTML` to `an HTML`.

cweitkamp (dismissed):
LGTM. Only minor remarks.

bundles/org.smarthomej.binding.mail/README.md
J-N-K: Good question. I don't know (both questions). I guess we should leave it like it is for the moment, I just noticed the missing space and added it.

cweitkamp (approved):

PR #202, opened 2021-10-13
[amazonechocontrol] refactoring and code improvements

J-N-K:
This PR refactors the connection process and simplifies code. The storage of the login data has been changed to JSON serialization instead of the previous proprietary code. This is backward compatible.


bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/Connection.java
```diff
@@ -1718,16 +1700,14 @@ private void queuedExecuteSequenceNode(QueueObject queueObject) {
...
-
             String text = executionNodeObject.text;
             if (text != null) {
                 text = text.replaceAll("<.+?>", " ").replaceAll("\\s+", " ").trim();
                 delay += text.length() * 150;
             }
 
-            makeRequest("POST", alexaServer + "/api/behaviors/preview", json, true, true, null, 3);
+            Map<String, String> headers = Map.of("Routines-Version", "1.1.218665");
+            makeRequest("POST", alexaServer + "/api/behaviors/preview", json, true, true, Map.of(), 3);
```
J-N-K: @Trinitus01 @mgeramb Any idea why we create a `Map` for custom headers and then don't use it?

Trinitus01: Seems this useless header goes back to the migration from 1.x to 2.x.
The getMusicProviders method is the only one which runs the makeRequest method using the Routines-Version header.
@mgeramb I think the header is not necessary at all, isn't it?

J-N-K:
Will be replaced by several smaller PR due to issues found while logging in.

PR #217, opened 2021-11-05
[transform.basicprofiles] Initial contribution of a time-range-command profile

cweitkamp:
- Initial contribution of a time-range-command profile (e.g. to implement a motion sensor)

Thoughts are welcome.


cweitkamp:
Ready for review.

J-N-K:
Thanks, code looks good. I'm not so happy with the naming (also I must admit that I have no good suggestion). The obvious use case is sending different brightness values at day and at night depending on detected motion. But the profile itself is not a motion sensor. WDYT about `TimedMappingProfile`?

bundles/org.smarthomej.transform.basicprofiles/README.md
```diff
@@ -145,3 +145,34 @@ This profile is a shortcut for the System Hysteresis Profile.
 ```java
 Switch thresholdItem { channel="xxx" [profile="basic-profiles:threshold", threshold=15] }
 ```
+
+## Motion Sensor Profile
+
+This is an enhanced implementation of a follow profile which converts `OnOffType` to a `PercentType`.
+The value of the percent type can be different between a specific time of the day (e.g. night).
```
J-N-K:
```suggestion
The value of the percent type can be different between a specific time of the day (e.g. night).
A possible use-case is switching lights (using a presence detector) with different intensities at day and at night.
```

cweitkamp:
Thanks for the feedback. Yes, that's true. The naming is not the best. What about `TimeBasedDimmingProfile` or `TimeRangeDimmingProfile`? Parameters should be renamed too.`dayValue`to `withinRangeValue` and `nightValue` to `outOfRangeValue` or similar.

J-N-K:
`TimeRangeValueProfile`? There could be other use-cases than the obvious dimming and it could be extended in the future to other types than PercentType. I agree on the parameter naming

cweitkamp:
I renamed it to `TimeRangeCommandProfile`.

J-N-K (dismissed):

J-N-K:
Can you apply spotless? the XML has format errors.

cweitkamp:
Done.

J-N-K (approved):

PR #254, opened 2021-12-24
[mail] fix thing actions not available from UI

J-N-K:
Reported on the forum. The reason is that the standard implementation uses a context class loader to get the mime type handlers and tries to load the handlers from there. Under some circumstances this is the class loader of the activation bundle which does doesn't have the `com.sun.mail` packages available, leading to an `javax.activation.UnsupportedDataTypeException: text/plain; charset=UTF-8` exception.

### DIFF
--- a/bundles/org.smarthomej.binding.mail/pom.xml
+++ b/bundles/org.smarthomej.binding.mail/pom.xml
@@ -39,4 +39,26 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>add-source</goal>
+            </goals>
+            <phase>generate-sources</phase>
+            <configuration>
+              <sources>
+                <source>src/3rdparty/java</source>
+              </sources>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+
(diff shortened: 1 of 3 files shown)


cweitkamp:
I am not sure if I understand the root cause for sure. What I can tell you is that I did not see a similar exception when using Jython scripting. Do you think it only occurs when using JS scripting?

Can it be related to https://github.com/openhab/openhab-addons/pull/11400?

J-N-K:
Not sure but since it also affects DSL rules (and only UI, not file-base rules) I doubt it.

The problem is that `javax.activation` contains the `MailcapCommandMap` which is used to select a mail-handler when sending a mail. When creating the object, it tries to load the required data via classloaders of several bundles. One of these bundles is `com.sun.mail/javax.mail/1.6.2`. Later, when sending the mail, a class name is looked up in the map (for the content-type) and the current threads ContextClassloader is used to load that class via reflection. It seems that in our case for UI defines rules the ContextClassloader doesn't contain the classes from `com.sun.mail/javax.mail/1.6.2`.

Since it worked before, it could also be that this is a result of the transition from `org.apache.servicemix.specs/org.apache.servicemix.specs.activation-api-1.1/2.9.0` to `org.apache.servicemix.specs/org.apache.servicemix.specs.activation-api-1.2.1/1.2.1_3` (even if the implementations looks the same for both versions).

cweitkamp (approved):
Alright, let's give it a try.

May add some comments to the patched class to mark the changes as a reminder for later.

J-N-K:
I added a "Changelog" to the file header

PR #259, opened 2021-12-27
[viessmann] fix refresh token expired after 180 day

rogrun:
This fixes the issue when the refresh token expires after 180 days. If refresh token is expired, then the full authorization process will be done. 

Some new features are added.

Signed-off-by: Ronny Grun [ronny.grun@t-online.de](mailto:ronny.grun@t-online.de)


bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/api/ViessmannAuth.java
```diff
@@ -209,6 +209,7 @@ private void getTokens() throws ViessmannAuthException {
         api.setTokenResponseDTO(tokenResponse);
         refreshToken = tokenResponse.refreshToken;
         api.setTokenExpiryDate(TimeUnit.SECONDS.toMillis(tokenResponse.expiresIn));
+        api.setRefreshTokenExpiryDate(TimeUnit.SECONDS.toMillis(REFRESH_TOKEN_EXPIRE));
```
J-N-K, line 212: Is this correct? Is the token always re-generated when the endpoint is called? Or could it be that an old token is retrieved that expires early?

rogrun: This is correct. When the endpoint is called, a new refresh token will be generated.

PR #260, opened 2021-12-28
[telenot] added UNKNOWN msgType

rogrun:
This PR adds UNKNOWN msgType. If the msgType isn't known, it will be logged as WARN. 
It also logs INVALID msgType as DEBUG and it will be ignored in the TelenotBridgeHandler.

Signed-off-by: Ronny Grun <ronny.grun@t-online.de>

### DIFF
--- a/bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/TelenotBridgeHandler.java
+++ b/bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/TelenotBridgeHandler.java
@@ -284,8 +284,14 @@ public abstract class TelenotBridgeHandler extends BaseBridgeHandler {
                             TelenotThingHandler.readyToSendData.set(true);
                             logger.trace("Ready to send data");
                             break;
-                        case INVALID:
+                        case UNKNOWN:
                             logger.warn("Received {} MsgType | hexString: {}", msgType, message);
+                            sendTelenotCommand(TelenotCommand.confirmACK());
+                            TelenotThingHandler.readyToSendData.set(true);
+                            logger.trace("Ready to send data");
+                            break;
+                        case INVALID:
+                            logger.debug("Received {} MsgType | hexString: {}", msgType, message);
                             sendTelenotCommand(TelenotCommand.confirmACK());
                             TelenotThingHandler.readyToSendData.set(true);
                             logger.trace("Ready to send data");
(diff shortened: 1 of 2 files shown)


bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/protocol/TelenotMsgType.java
```diff
@@ -120,6 +121,12 @@ public static TelenotMsgType getMsgType(@Nullable String s) {
...
 
+        if (mt == null) {
+            if (s.matches("^6868(.*)16")) {
+                mt = TelenotMsgType.UNKNOWN;
+            }
+        }
+
         if (mt == null) {
             mt = TelenotMsgType.INVALID;
         }
```
J-N-K:
```suggestion
        if (mt == null) {
            if (s.matches("^6868(.*)16")) {
                mt = TelenotMsgType.UNKNOWN;
            } else {
                mt = TelenotMsgType.INVALID;
            }
        }
```

bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/TelenotBridgeHandler.java
```diff
@@ -284,12 +284,18 @@ private void readerThread() {
                             TelenotThingHandler.readyToSendData.set(true);
                             logger.trace("Ready to send data");
                             break;
-                        case INVALID:
+                        case UNKNOWN:
```
J-N-K, line 287: What is the difference between INVALID and UNKNOWN? I'm asking because one is logged at WARN and the other at DEBUG. What is the user supposed to do when he encounters an UNKNOWN message type?

J-N-K (changes requested):

bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/TelenotBridgeHandler.java
rogrun: Today I updated my live system to oh3.2
At the start I received an incomplete message (INVALID)
Every valid message begins with "6868" and ends with "16"
This invalid message did not contain it.

However, some messages (even valid ones) may be unknown if the msgType is not currently implemented.

I hope you are okay with this.

J-N-K: The question is why you received invalid/incomplete messages. This should be investigated. I'll merge that here anyway, sounds good.

J-N-K (approved):

PR #286, opened 2022-01-03
[deconz] add externalwindowopen channel to thermostat

J-N-K:

### DIFF
--- a/bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/thing/sensor-thing-types.xml
+++ b/bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/thing/sensor-thing-types.xml
@@ -560,7 +560,7 @@
 		<label>Window Open</label>
 	</channel-type>
 	<channel-type id="externalwindowopen">
-		<item-type>Switch</item-type>
+		<item-type>Contact</item-type>
 		<label>External Window Open</label>
 	</channel-type>
 	<channel-type id="heatsetpoint">
--- a/bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/thing/sensor-thing-types.xml
+++ b/bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/thing/sensor-thing-types.xml
@@ -555,12 +555,20 @@
 		<description>Status of this thermostat's childlock.</description>
 		<category>Lock</category>
 	</channel-type>
+	<channel-type id="windowopen">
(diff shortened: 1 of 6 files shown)


bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/thing/sensor-thing-types.xml
```diff
@@ -555,12 +555,20 @@
 		<description>Status of this thermostat's childlock.</description>
 		<category>Lock</category>
 	</channel-type>
+	<channel-type id="windowopen">
+		<item-type>Contact</item-type>
+		<label>Window Open</label>
+	</channel-type>
+	<channel-type id="externalwindowopen">
+		<item-type>Switch</item-type>
```
J-N-K: @cweitkamp I'm unhappy with this modeling as a "Window Open" usually is a `Contact`. But `OpenClosedType` is not a command and this is an input to an external device (it allows to close the thermostat's valve in case the window is open), so we need a command. Or do you think this is a corner-case where the thing handler should implement `handleUpdate`?

cweitkamp: Isn't `String` an option? With Command/StateOptions? This way you can even say that the external window can have more than two states like open, close and tilt.

J-N-K: For the API it's mapped to a boolean, so there are only two possible states. We already have a `windowopen` status (for thermostats which have a direct-connected window contact) channel which uses `Contact` as item-type (like `open` channel for Open/Closed sensors). All other boolean states are `Switch` channels. I fear that having a `String` channel here is even more confusing than the `Switch`.

J-N-K: The more I think about it, the more I think that `handleUpdate` would be the correct solution. What is essentially expected here is to pass a state update from a window contact to the thermostat. It's not a command but really a state.

J-N-K: Forget my comment. I just noticed we don't have `handleUpdate` any longer. Somehow I remembered that `OpenCloseType` didn't implement `Command` (like `RawType`) but in fact it does. So a `Contact` is perfectly fine with using the Follow profile

PR #293, opened 2022-01-07
[viessmann] add uom and consumption channels

rogrun:
This PR includes a refactoring for UOM and adding channels for consumption.

I have also added the Viessmann logo (Viessmann allowed me to do this / I asked for it)
Is it possible to add this logo in the bindings area on OH like on the official bindings?


J-N-K:
Image: Not currently, because the addon-service is not providing an image to openHAB, but I'll raise a PR for that.

Regarding the UoM change: I didn't have time to check the code yet (hope to find time tomorrow). Are channels that were created with an older version of the binding automatically changed to the new channel-types?

rogrun:
Image: Sounds good. 😀

UoM: The existing channels are not automatically changed to the new channel-types. Currently the thing has to be deleted and rediscovered.

Is there a way to do this automatically? 

Do not hurry, it would be nice if it is in the new release.

J-N-K:
If I remember correctly, channels are added dynamically when an update is received and the channel is missing, is that correct? If so, it's quite easy to do. How many channels are affected?

rogrun:
Yes, that's right, channels are added dynamically. About 30 channels are affected.

J-N-K:
Ok, it's a bit work, but you need to add a dependency

```
  <dependencies>
    <dependency>
      <groupId>org.smarthomej.addons.bundles</groupId>
      <artifactId>org.smarthomej.commons</artifactId>
      <version>${project.version}</version>
      <scope>provided</scope>
    </dependency>
  </dependencies>
```

then inherit the thing handler from `UpdatingBaseThingHandler` instead of `BaseThingHandler`.

Then create a file `update/device.update` with the content

```
1;REMOVE_CHANNEL;channel1
1;REMOVE_CHANNEL;channel2
1;REMOVE_CHANNEL;channel3
...
```

and finally add a property `<thingTypeVersion>1</thingTypeVersion>` to the `device`-thing (in the `OH-INF/things/thing-types.xml`.

rogrun:
All changes are done. But the `thingTypeVersion` property is not automatically added to the `device thing`
Only deletion and rediscovery works. Is it correct that way?

J-N-K:
No. It should be added and it works for me. Are you sure it is not added? Do you see the `Updating {} from version {} to {}` message again on next startup?

rogrun:
I have made a mistake. I created the update folder in the wrong place. Now it works for me. Thanks :)

rogrun:
Everything is done. Ready for review.

bundles/org.smarthomej.binding.viessmann/README.md
```diff
@@ -1,15 +1,19 @@
 # Viessmann Binding
 
+<img src="/bundles/org.smarthomej.binding.viessmann/doc/viessmann_wordmark_rgb_1_vitorange.png" width="140"/>
```
J-N-K: I doubt that works on the documentation website. I would suggest to change it to `org.smarthomej.binding.viessmann/doc/viessmann_wordmark_rgb_1_vitorange.png`, then I can try to add that to the website.

bundles/org.smarthomej.binding.viessmann/README.md
```diff
@@ -1,15 +1,19 @@
...
 
 ## Note / Important
 
-You have to register your ViCare Account at the Viessmann developer portal and create a Client ID.
+You have to register your ViCare Account at the [Viessmann developer portal](https://developer.viessmann.com/) and create a Client ID.
 
 * `name` - `i.e. openhab`
 * `Google reCAPTCHA` - `off`
-* `Redicect URI` - `http://localhost:8080/viessmann/authcode/`
+* `Redicect URI` - `http://localhost:8080/viessmann/authcode/` (*)
```
J-N-K:
```suggestion
* `Redirect URI` - `http://localhost:8080/viessmann/authcode/` (*)
```

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/api/ViessmannAuth.java
```diff
@@ -134,49 +136,53 @@ public void setAccessToken(String newAccessToken) {
...
-            } else {
-                AuthorizeResponseDTO resp = this.authResponse;
-                if (resp == null) {
-                    logger.warn("AuthorizeResponseDTO is null. This should not happen.");
+        if (callbackUrl != null) {
+            StringBuilder url = new StringBuilder(VIESSMANN_AUTHORIZE_URL);
+            url.append("?response_type=code");
+            url.append("&client_id=").append(apiKey);
+            url.append("&code_challenge=2e21faa1-db2c-4d0b-a10f-575fd372bc8c-575fd372bc8c");
+            url.append("&redirect_uri=" + callbackUrl + "/viessmann/authcode/");
```
J-N-K:
```suggestion
            url.append("&redirect_uri=").append(callbackUrl).append("/viessmann/authcode/");
```

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/api/ViessmannAuth.java
```diff
@@ -134,49 +136,53 @@ public void setAccessToken(String newAccessToken) {
...
+            url.append("&scope=").append(VIESSMANN_SCOPE);
+            logger.trace("ViessmannAuth: Getting authorize URL={}", url);
+            String response = executeUrlAuthorize("GET", url.toString());
+            logger.trace("ViessmannAuth: Auth response: {}", response);
+            if (response != null) {
+                if (response.indexOf("<!DOCTYPE html>") >= 0) {
```
J-N-K:
```suggestion
                if (response.contains("<!DOCTYPE html>")) {
```

also below. While browsing the existing code I've seen some 
`var.indexOf(...) == -1`. `!var.contains(...)` is better style.

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/api/ViessmannAuth.java
```diff
@@ -134,49 +136,53 @@ public void setAccessToken(String newAccessToken) {
...
+                if (response.indexOf("error") >= 0) {
+                    logger.warn("ViessmannAuth: Login failed. Wrong code response.");
                     return;
                 }
-                if (resp.errorMsg != null) {
+            }
+            try {
+                authResponse = api.getGson().fromJson(response, AuthorizeResponseDTO.class);
+                if (authResponse == null) {
                     logger.debug("ViessmannAuth: Got null authorize response from Viessmann API");
```
J-N-K, line 163: Just a comment (no need to change): `Json.fromJson` only returns `null` if the input is `null`. This is clearly not the case here. Unfortunately out null-checker can't detect that, so we have to help him. Either by the "classic" null-check (what is your code) or by enforcing non-null:

`authReponse = Objects.requireNonNull(api.getGson.fromJson(response, AuthorizeResponseDTO.class);`

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/dto/UnitsMap.java
```diff
@@ -0,0 +1,53 @@
...
+import org.openhab.core.library.unit.Units;
+
+/**
+ * This class provides the units for values depending on the viessmann unit
+ * 
+ * @author Ronny Grun - Initial contribution
+ */
+@NonNullByDefault
+public class UnitsMap {
+    private static final Map<String, String> UNIT_MAP = new HashMap<>();
```
J-N-K:
```suggestion
    private static final Map<String, String> UNIT_MAP = Map.of( //
      "celsius", SIUnits.Celsius.getSymbol(), //
      ...
    );
```

Then you can omit the `static` part below. You can also make the `Map` public and add it to the binding constants class. No need to add a new class for that.

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/handler/DeviceHandler.java
```diff
@@ -226,169 +230,253 @@ public void handleUpdate(FeatureDataDTO featureDataDTO) {
...
-
-                            if (thing.getChannel(subMsg.getChannelId()) == null && !"unit".equals(entry)) {
-                                createSubChannel(subMsg);
+
+                    if (msg.getDeviceId().indexOf(config.deviceId) != -1 && !"unit".equals(entry)) {
+                        if (!"[]".equals(valueEntry)) {
+                            if (viUnit != null) {
+                                if (!viUnit.isEmpty()) {
+                                    msg.setUnit(viUnit);
+                                    unit = UnitsMap.getUnit(viUnit);
```
J-N-K: If you make the units map a constant, you can just use .get here

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/handler/DeviceHandler.java
```diff
@@ -226,169 +230,253 @@ public void handleUpdate(FeatureDataDTO featureDataDTO) {
...
+                                            updateState(msg.getChannelId(), state);
+                                        } else {
+                                            updateState(msg.getChannelId(),
+                                                    new QuantityType<>(msg.getValue() + " " + unit));
+                                        }
+                                    }
+                                    break;
+                                case "percent":
+                                    updateState(msg.getChannelId(),
+                                            new QuantityType<>(Double.valueOf(msg.getValue()), Units.PERCENT));
```
J-N-K: Also below: Why don't you use the unit map for that, too? Then you can just have one code for that instead of many cases.

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/handler/DeviceHandler.java
```diff
@@ -226,169 +230,253 @@ public void handleUpdate(FeatureDataDTO featureDataDTO) {
...
-                            updateState(msg.getChannelId(), state);
-                            String channelId = msg.getChannelId().replace("#schedule", "#produced");
-                            updateState(channelId, parseSchedule(msg.getValue()));
+                            switch (typeEntry) {
+                                case "decimal":
+                                case "number":
+                                case "temperature":
+                                    if (unit != null) {
+                                        if (unit.isEmpty()) {
+                                            DecimalType state = DecimalType.valueOf(msg.getValue());
```
J-N-K: Wouldn't this also be the correct handling for `unit==null`?

J-N-K (changes requested):
Thanks for your contribution. Good job and nice additions. I have left some comments inside.

bundles/org.smarthomej.binding.viessmann/README.md
rogrun: I changed it, but now it no longer works on github. Let's see what happens on the doc website.

rogrun:
Thanks :) 
I have done the requested changes.

Good to know with `Json.fromJson` :)

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/handler/DeviceHandler.java
```diff
@@ -226,169 +228,251 @@ public void handleUpdate(FeatureDataDTO featureDataDTO) {
...
+                                            new QuantityType<>(Double.valueOf(msg.getValue()) + " " + unit));
+                                    break;
+                                case "minute":
+                                    updateState(msg.getChannelId(),
+                                            new QuantityType<>(Double.valueOf(msg.getValue()) + " " + unit));
+                                    break;
+                                case "hours":
+                                    updateState(msg.getChannelId(),
+                                            new QuantityType<>(Double.valueOf(msg.getValue()) + " " + unit));
+                                    break;
```
J-N-K:
```suggestion
                                case "percent":
                                case "minute":
                                case "hours:
                                    updateState(msg.getChannelId(),
                                            new QuantityType<>(Double.valueOf(msg.getValue()) + " " + unit));
                                    break;
```

J-N-K (changes requested):
Just one last comment

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/handler/DeviceHandler.java
rogrun: Done

J-N-K (approved):

PR #295, opened 2022-01-13
[deconz] fix websocket re-connection

J-N-K:
Under some circumstances a websocket connection was started twice when trying to recover after a lost connection. Reported on Facebook.
 
### DIFF
--- a/bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/netutils/WebSocketConnection.java
+++ b/bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/netutils/WebSocketConnection.java
@@ -56,7 +56,7 @@ public class WebSocketConnection {
     private final WebSocketClient client;
     private final String socketName;
     private final Gson gson;
-    private final int watchdogInterval;
+    private int watchdogInterval;
 
     private final WebSocketConnectionListener connectionListener;
     private final Map<String, WebSocketMessageListener> listeners = new ConcurrentHashMap<>();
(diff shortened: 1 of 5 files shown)


bundles/org.smarthomej.binding.deconz/src/main/java/org/smarthomej/binding/deconz/internal/netutils/WebSocketConnection.java
```diff
@@ -76,6 +76,10 @@ public WebSocketConnection(WebSocketConnectionListener listener, WebSocketClient
         this.watchdogInterval = watchdogInterval;
     }
 
+    public void setWatchdogInterval(int watchdogInterval) {
+        this.watchdogInterval = watchdogInterval;
+    }
+
     public void start(String ip) {
```
cweitkamp, line 83: I am missing a line of code which sets the `connectionState` to `ConnectionState.CONNECTING`. Do you think that is the reason for multiple tries to start the websocket?

J-N-K: Yes, that's the reason why it tries to start multiple times. I fixed it.

Nevertheless the multiple calls to `connectionListener.webSocketConnectionLost(reason)` would still create multiple new instances on a single connection loss (and probably leak memory). This is the result of the call in l. 101/106 is that sometimes the `@OnWebsocketClose` is not called, it depends on when the connection attempt fails. If it is called, the bridge handler is notified two times. If we omit the call here, we could miss some failed connection attempts. Unfortunately it is not clearly documented when `@OnWebsocketClose` is called.

bundles/org.smarthomej.binding.deconz/src/main/resources/OH-INF/config/config.xml
```diff
@@ -32,6 +32,12 @@
 			<advanced>true</advanced>
 			<default>2000</default>
 		</parameter>
+		<parameter name="websocketTimeout" type="integer" unit="s" min="30">
+			<label>Websocket Idle Timeout</label>
```
cweitkamp: Should we label the two timeouts in "HTTP Timeout" and "Websocket Timeout"? Similar to the port.

Alternatively we can introduce parameter-groups.

cweitkamp (dismissed):
Alright.. LGTM.

One final comment regarding design.

cweitkamp (approved):

PR #300, opened 2022-01-17
[telenot] fix channel naming with umlauts and fix reconnection

rogrun:
This PR contains following fixes:

1. Channel naming with umlauts have the wrong char. Telenot provides the wrong hex value for the charset. 
2. Some RS232 to Ethernet needs a delay between disconnect and reconnect.

### DIFF
--- a/bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/IPBridgeHandler.java
+++ b/bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/IPBridgeHandler.java
@@ -44,6 +44,8 @@ public class IPBridgeHandler extends TelenotBridgeHandler {
 
     private @Nullable Socket socket = null;
 
+    private @Nullable ScheduledFuture<?> connectJob = null;
+
     public IPBridgeHandler(Bridge bridge) {
         super(bridge);
     }
@@ -71,7 +73,14 @@ public class IPBridgeHandler extends TelenotBridgeHandler {
 
     @Override
     protected synchronized void connect() {
-        disconnect(); // make sure we are disconnected
+        disconnect();
+        ScheduledFuture<?> connectJob = this.connectJob;
+        if (connectJob == null || connectJob.isDone()) {
+            connectJob = scheduler.schedule(this::internalConnect, 2, TimeUnit.SECONDS);
+        }
+    }
+
+    protected synchronized void internalConnect() {
         writeException = false;
         try {
             Socket socket = new Socket(config.hostname, config.tcpPort);
@@ -176,16 +185,20 @@ public class IPBridgeHandler extends TelenotBridgeHandler {
             updateTelenotClockJob = null;
         }
 
+        ScheduledFuture<?> connectJob = this.connectJob;
+        if (connectJob != null) {
+            // use cancel(false) so we don't kill ourselves when reconnect job calls disconnect()
+            connectJob.cancel(false);
+            connectJob = null;
+        }
+
         // Must close the socket first so the message reader thread will exit properly.
         Socket s = socket;
         if (s != null) {
             try {
                 s.close();
-                Thread.sleep(2000);
             } catch (IOException e) {
                 logger.debug("error closing socket: {}", e.getMessage());
-            } catch (InterruptedException e) {
-                logger.debug("waiting after closing socket fails: {}", e.getMessage());
             }
         }
         socket = null;
--- a/bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/IPBridgeHandler.java
+++ b/bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/IPBridgeHandler.java
@@ -181,8 +181,11 @@ public class IPBridgeHandler extends TelenotBridgeHandler {
         if (s != null) {
             try {
                 s.close();
+                Thread.sleep(2000);
             } catch (IOException e) {
                 logger.debug("error closing socket: {}", e.getMessage());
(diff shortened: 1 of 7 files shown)


bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/IPBridgeHandler.java
```diff
@@ -181,8 +181,11 @@ protected synchronized void disconnect() {
         if (s != null) {
             try {
                 s.close();
+                Thread.sleep(2000);
```
J-N-K: Why do you add a sleep here? In general sleep should be avoided.

bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/TelenotBridgeHandler.java
```diff
@@ -303,6 +304,10 @@ private void readerThread() {
                 } catch (MessageParseException e) {
                     logger.warn("Error {} while parsing message {}. Please report bug.", e.getMessage(), message);
                 }
+                if (discoveryStarted && usedInputContact.isEmpty() && usedReportingArea.isEmpty()) {
+                    discoveryStarted = false;
+                    logger.info("Discovery job completed");
```
J-N-K, line 309: Reduce to debug

bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/protocol/UsedContactInfoMessage.java
```diff
@@ -50,8 +50,22 @@ public UsedContactInfoMessage(String message) throws TelenotMessageException {
         int contactHexLen = contactNameHex.length();
         int b = 0;
         while (b < contactHexLen) {
-            Integer charcode = Integer.parseInt(contactNameHex.substring(b, b + 2), 16);
-            strBuilder.append(String.valueOf(Character.toChars(charcode)));
+            String hex = contactNameHex.substring(b, b + 2);
```
J-N-K: What do you want to achieve here? If I understand the code correct, you have an input string which is a hex-encoded byte-array that contains the name encoded in LATIN-1, so your input is `54FC72` and it should be transformed to `Tür`. This can be achieved in one line by 

```
String strcontact = new String(HexUtils.hexToBytes(contactNameHex), StandardCharsets.ISO_8859_1);
```

J-N-K (changes requested):

bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/IPBridgeHandler.java
rogrun: I want to wait 2 seconds before the reconnect. Can you help for a better way?

bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/TelenotBridgeHandler.java
rogrun: I want to show the user that the discovery is in progress. It takes about 5 minutes to discover all channels, so it would be nice to know when it will be completed.

bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/IPBridgeHandler.java
J-N-K: I guess this issue is only relevant for the call to `disconnect` and the beginning of `connect`, is that right? Then you could refactor connect to

```
public void connect() {
    disconnect();
    ScheduledFuture<> connectJob = this.connectJob;
    if (connectJob == null || connectJob.isDone()) { 
        connectJob = scheduler.schedule(this::internalConnect, 2, TimeUnit.SECONDS);
    }
}
```

and move the rest of the code to `internalConnect()`. Make sure that `connectJob` is also cancelled on disposal.

bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/TelenotBridgeHandler.java
J-N-K: Ok, then leave it at INFO.

bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/protocol/UsedContactInfoMessage.java
rogrun: The alarm system returns the wrong hex for umlauts, so I get the wrong character.
The input for `Tür` is `54F572`, but it converts to `Tõr` 
Because `F5` is `õ`
Maybe i could replace all wrong hex to the right and then converts the hex-encoded byte-array to string

J-N-K: Oh. I didn't notice that they use the weird HD44780 encoding. 

I guess that

```
String strcontact = new String(HexUtils.hexToBytes(contactNameHex), StandardCharsets.ISO_8859_1).replace("õ", "ü").replace(...);

```

is still a bit more clear.

rogrun:
Thanks for your support. All changes are done.

J-N-K (approved):

PR #312, opened 2022-01-30
[telenot] add handling for unused contact

rogrun:
Some installations discover contacts that are unused. In this case, the discovery runs in a loop.

Signed-off-by: Ronny Grun <ronny.grun@t-online.de>


bundles/org.smarthomej.binding.telenot/src/main/java/org/smarthomej/binding/telenot/internal/handler/TelenotBridgeHandler.java
```diff
@@ -298,6 +298,16 @@ private void readerThread() {
                             TelenotThingHandler.readyToSendData.set(true);
                             logger.trace("Ready to send data");
                             break;
+                        case NOT_USED_CONTACT:
+                            logger.debug("Received {} MsgType | hexString: {}", msgType, message);
+                            if (!usedInputContact.isEmpty()) {
+                                logger.warn("Contact {} not used", usedInputContact.get(0));
```
J-N-K: Is this an error state or is this expected? If it's not an error, please reduce to INFO.

rogrun: It is expected. I changed this to INFO

J-N-K:
Please run spotless. Thanks.

PR #328, opened 2022-02-21
[tuya] Fix a typo on SilentMode

jimtng:
This is a simple typo correction. HOWEVER, I think this channel, at least on my heat pump, has the opposite meaning. Instead of "Silent Mode", it should be called "Boost".

When it's ON, it goes full power 100% which is the opposite of silent. When it's OFF, it limits the maximum power to 80% to make it quieter.

Maybe on some other models it works for "Silent mode" / night mode

### DIFF
--- a/bundles/org.smarthomej.binding.tuya/src/main/resources/schema.json
+++ b/bundles/org.smarthomej.binding.tuya/src/main/resources/schema.json
@@ -21575,7 +21575,7 @@
 			"id": 116,
 			"type": "bitmap"
 		},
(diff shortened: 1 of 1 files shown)


J-N-K:
Thanks. I hope I keep this in mind when updating this file. It is auto-generated from https://github.com/Apollon77/ioBroker.tuya/blob/master/lib/schema.json and this file is auto-generated from a database (which probably contains the typo).

J-N-K (approved):

PR #378, opened 2022-07-02
[mail] Fix sending HTML/Multipart mails

J-N-K:
The fix in #254 was not sufficient for multipart messages as they contain a `DataHandler` for each body part. The command map was only set for the outer mime part, not for the body parts. This is not a nice solution as it relies on internal implementation of other classes but it is not accessible otherwise.


J-N-K:
@hmerk You might be interested in this one.

hmerk:
Thanks @J-N-K, will port this to openHAB-addons as we did for the other fix.

PR #401, opened 2022-09-04
[androiddebugbridge] Remove package prefix while determination of current running package

cweitkamp:
- Remove package prefix while determination of current running package (in fallback method)
- fix `channelFallbackMap`

Related to #396


J-N-K (commented):
LGTM, but it seems spotless does not like it. Can you re-apply formatting?

cweitkamp:
Fixed in the same second.

J-N-K (approved):

PR #403, opened 2022-10-02
[viessmann] add command option and state description

rogrun:
This PR adds automatically generated command options depending on the provided commands from the API


cweitkamp:
One comment. You do not need both a `CommandDescription` and a `StateDescription` provider for the same channel. If you have a channel exposing a state use the later one. It decodes states to commands with the same meaning. If you have a state-less Channel which use the `CommandDescriptionProvider`.

rogrun:
Good to know, I will change that and give feedback on it

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/ViessmannDynamicCommandDescriptionProvider.java
```diff
@@ -0,0 +1,42 @@
...
+import org.osgi.service.component.annotations.Reference;
+
+/**
+ * Dynamic provider of command options.
+ *
+ * @author Ronny Grun - Initial contribution
+ */
+@Component(service = { DynamicCommandDescriptionProvider.class, ViessmannDynamicCommandDescriptionProvider.class })
+@NonNullByDefault
+public class ViessmannDynamicCommandDescriptionProvider extends BaseDynamicCommandDescriptionProvider {
```
cweitkamp: Please correct me if I am wrong: you can remove the entire class.

cweitkamp (changes requested):

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/ViessmannDynamicCommandDescriptionProvider.java
rogrun: yes of course :) I'll remove this class

bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/handler/DeviceHandler.java
```diff
@@ -60,6 +63,7 @@
  * @author Ronny Grun - Initial contribution
  */
 @NonNullByDefault
+@SuppressWarnings("null")
```
J-N-K: Why do you suppress this warning?

rogrun: I use VS Code as an editor. Recently I've been getting warnings like this one:
`Null type safety (type annotations): The expression of type 'Logger' needs unchecked conversion to conform to '@NonNull Logger'`
I solved it by suppress this. I don't know if it is the right way. Should I change this?

cweitkamp (dismissed):

cweitkamp (approved):

J-N-K (approved):

PR #429, opened 2022-12-09
[amazonechocontrol] Fix sequence ignores standard volume

Trinitus01:
fixed: sequence ignores standard volume

### DIFF
--- a/bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/connection/Connection.java
+++ b/bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/connection/Connection.java
@@ -1378,11 +1378,12 @@ public class Connection {
             Map<String, Object> parameters, List<@Nullable Integer> ttsVolumes,
             List<@Nullable Integer> standardVolumes) {
         JsonArray serialNodesToExecute = new JsonArray();
+
         JsonArray ttsVolumeNodesToExecute = new JsonArray();
         for (int i = 0; i < devices.size(); i++) {
             Integer ttsVolume = ttsVolumes.size() > i ? ttsVolumes.get(i) : null;
             Integer standardVolume = standardVolumes.size() > i ? standardVolumes.get(i) : null;
-            if (ttsVolume != null && (standardVolume != null || !ttsVolume.equals(standardVolume))) {
+            if (ttsVolume != null && !ttsVolume.equals(standardVolume)) {
                 ttsVolumeNodesToExecute.add(
                         createExecutionNode(devices.get(i), "Alexa.DeviceControls.Volume", Map.of("value", ttsVolume)));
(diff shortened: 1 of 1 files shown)


bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/connection/Connection.java
```diff
@@ -1413,9 +1414,9 @@ private void executeSequenceCommandWithVolume(List<Device> devices, @Nullable St
 
         JsonArray standardVolumeNodesToExecute = new JsonArray();
         for (int i = 0; i < devices.size(); i++) {
-            Integer ttsVolume = ttsVolumes.size() > i ? ttsVolumes.get(i) : null;
             Integer standardVolume = standardVolumes.size() > i ? standardVolumes.get(i) : null;
-            if (ttsVolume != null && standardVolume != null && !ttsVolume.equals(standardVolume)) {
+            Integer ttsVolume = ttsVolumes.size() > i ? ttsVolumes.get(i) : null;
+            if (ttsVolume != null && standardVolume != null && !standardVolume.equals(ttsVolume)) {
```
J-N-K, line 1419:
```suggestion
            if (standardVolume != null && !standardVolume.equals(ttsVolume)) {
```

`.equals` is null-safe

J-N-K (changes requested):
Otherwise LGTM

bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/connection/Connection.java
Trinitus01: Yes but if we remove the null check of ttsVolume, it will reset the volume to standardVolume although it is the same as ttsVolume. For example standardVolume is 50 and ttsVolume is 50, it will not set the volume, but it would reset the volume. So we need to check if ttsVolume is set and if not, we also should not reset the volume.

Trinitus01: Because if we set or reset the volume and ttsVolume and standardVolume are equals, the Echo will still output a peep sound.

Trinitus01: The ttsVolume null check is for: "Do only reset the volume if we also set it"

J-N-K: I just realized that when answering to your last comment....

J-N-K (approved):

PR #433, opened 2022-12-20
[viessmann] adds the ability to manually renew all device channels via console command

rogrun:
@cweitkamp I implemented the ability to manually renew all device channels via console command. What do you think about it?


J-N-K:
Isn't that essentially the same as deleting and re-adding the thing?

rogrun:
Yes, it is. I would like to simplify it in this way and give the user the opportunity to do it via the console. I'm not sure if this fits the concept of OpenHAB. If not, I can dismiss this idea.

rogrun:
@J-N-K and @cweitkamp WDYT about this PR?

Issue #488, opened 2023-05-27
Divide profile handling of NULL


clinique:
I quite often find the following in my logs : 

`2023-05-27 07:20:08.602 [WARN ] [profiles.DivideTransformationProfile] - Could not apply 'DIVIDE' transformation on state 'NULL' with value '8'
`
I propose to test if sourceString is NULL prior to try the conversion 
[here :](https://github.com/smarthomej/addons/blob/3.2.x/bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java)

...and maybe return NULL also. Your thoughts ?

PR #489, opened 2023-05-30
Null handling in divide

clinique:
This is my proposal to address what I think is an issue #488

### DIFF
--- a/bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
+++ b/bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
@@ -34,26 +34,28 @@ abstract class AbstractMathTransformationService implements TransformationServic
 
     @Override
     public @Nullable String transform(String valueString, String sourceString) throws TransformationException {
-        QuantityType<?> source;
-        try {
-            source = new QuantityType<>(sourceString);
-        } catch (IllegalArgumentException e) {
-            logger.warn("Input value '{}' could not be converted to a valid number", sourceString);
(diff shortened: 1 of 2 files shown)


bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
```diff
@@ -34,27 +34,29 @@ abstract class AbstractMathTransformationService implements TransformationServic
...
-            throw new TransformationException("Math Transformation can only be used with numeric inputs");
-        }
-        QuantityType<?> value;
-        try {
-            value = new QuantityType<>(valueString);
-        } catch (IllegalArgumentException e) {
-            logger.warn("Input value '{}' could not be converted to a valid number", valueString);
-            throw new TransformationException("Math Transformation can only be used with numeric inputs");
+        
+        if("NULL".equals(valueString) || "NULL".equals(sourceString)) {
```
J-N-K, line 38: Wouldn't it make sense to check `UNDEF`, too?

J-N-K: @clinique Can you comment here? I would like to do a release in the next days due to #490.

clinique: > Wouldn't it make sense to check `UNDEF`, too?

You're correct, I thought of it but I only faced the error with NULL, never UNDEF, so thought : if not broken, don't fix it :)

J-N-K:
I added another version to get it into 3.2.17

PR #492, opened 2023-06-04
[viessmann] fix npe error reading

rogrun:
This PR fixes a NPE at error reading and out of bounds error on handle error messages


bundles/org.smarthomej.binding.viessmann/src/main/java/org/smarthomej/binding/viessmann/internal/handler/ViessmannBridgeHandler.java
```diff
@@ -219,10 +219,12 @@ public void getDeviceError() {
             countApiCalls();
             logger.trace("Errors:{}", errors);
             if (errors != null) {
-                String state = errors.data.get(0).body.errorDescription;
-                Boolean active = errors.data.get(0).body.active;
-                updateState("lastErrorMessage", StringType.valueOf(state));
-                updateState("errorIsActive", OnOffType.from(active));
+                if (errors.data.size() > 0) {
```
J-N-K: You can collapse this statement:

```
if (errors != null && errors.data.size() > 0)
```

This is safe because expressions are evaluated from left to right and aborted if `errors == null`.

J-N-K (changes requested):

J-N-K (approved):

PR #507, opened 2023-07-24
[transform.math] Fix missing unit when result is 0

### LINKED ISSUE #506: [Math transform] Unit is ignored if result is 0
At night I observe this warning : 

`2023-07-21 22:24:54.597 [WARN ] [penhab.core.library.items.NumberItem] - Failed to update item 'Total_Radiation_COR' because '0' could not be converted to the item unit 'W/m²'`

The item definition is : 

```
Number:Intensity Total_Radiation_COR "Radiation pondérée" <radiation> (gAstroSun1,gSynopanalyzer1) ["Point"] {unit="W/m²", channel="astro:sun:1:radiation#total" [profile="transform:MULTIPLY", multiplicand=0, itemName="Synop_Attenuation"]}
```

It takes current solar radiation : 

```
Number:Intensity Total_Radiation "Radiation Totale" <radiation> (gAstroSun1) ["Point"] {unit="W/m²", channel="astro:sun:1:radiation#total"}
```
and multiplies it by the cloudiness factor (that has no unit) : 

```
Number Synop_Attenuation "Coefficient Atténuation" <sun_clouds> (gSynopanalyzer1) ["Measurement"] {channel="synopanalyzer:synopanalyzer:1:attenuation-factor"}
```

During the day, it works fine but at night Total_Radiation = 0 W/m² and the warning is emitted.
Taking a look at `AbstractMathTransformationService`, I found the origin of the problem : 

```
QuantityType<?> result = performCalculation(source, value);
return BigDecimal.ZERO.compareTo(result.toBigDecimal()) == 0 ? "0" : result.toString();
```
So, if the result is 0, the code drops the unit, leading to the warning.

I think this could be changed to : 

`
return performCalculation(source, value).toString();
`

@J-N-K : what do you think ?

clinique:
Resolves #506

### DIFF
--- a/bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
+++ b/bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
@@ -12,6 +12,8 @@
  */
 package org.smarthomej.transform.math.internal;
 
+import java.math.BigDecimal;
+
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.library.types.QuantityType;
@@ -39,18 +41,21 @@ abstract class AbstractMathTransformationService implements TransformationServic
         }
 
         QuantityType<?> source;
-        QuantityType<?> value;
-        String evaluated = sourceString;
         try {
-            source = new QuantityType<>(evaluated);
-            evaluated = valueString;
-            value = new QuantityType<>(evaluated);
+            source = new QuantityType<>(sourceString);
         } catch (IllegalArgumentException e) {
-            logger.warn("Input value '{}' could not be converted to a valid number", evaluated);
+            logger.warn("Input value '{}' could not be converted to a valid number", sourceString);
             throw new TransformationException("Math Transformation can only be used with numeric inputs");
         }
-        
+        QuantityType<?> value;
         try {
+            value = new QuantityType<>(valueString);
+        } catch (IllegalArgumentException e) {
+            logger.warn("Input value '{}' could not be converted to a valid number", valueString);
+            throw new TransformationException("Math Transformation can only be used with numeric inputs");
+        }
+        try {
+            QuantityType<?> result = performCalculation(source, value);
             return performCalculation(source, value).toString();
         } catch (IllegalArgumentException e) {
             throw new TransformationException("ArithmeticException: " + e.getMessage());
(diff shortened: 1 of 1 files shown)


bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
```diff
@@ -41,22 +39,19 @@ abstract class AbstractMathTransformationService implements TransformationServic
...
         try {
-            value = new QuantityType<>(valueString);
+            source = new QuantityType<>(evaluated);
+            evaluated = valueString;
+            value = new QuantityType<>(evaluated);
         } catch (IllegalArgumentException e) {
-            logger.warn("Input value '{}' could not be converted to a valid number", valueString);
+            logger.warn("Input value '{}' could not be converted to a valid number", evaluated);
             throw new TransformationException("Math Transformation can only be used with numeric inputs");
         }
```
J-N-K: Why did you change that?

bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
```diff
@@ -41,22 +39,19 @@ abstract class AbstractMathTransformationService implements TransformationServic
...
+        
         try {
-            QuantityType<?> result = performCalculation(source, value);
-            return BigDecimal.ZERO.compareTo(result.toBigDecimal()) == 0 ? "0" : result.toString();
+            return performCalculation(source, value).toString();
```
J-N-K: I See why it works now, but I wonder why it was done like that before. Maybe to prevent a return value of `0.0000`?

bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
clinique: To avoid having two catch catching the same thing....could be avoided.

bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
clinique: Yes, I was wondering the same, reason why I wanted your advice.

bundles/org.smarthomej.transform.math/src/main/java/org/smarthomej/transform/math/internal/AbstractMathTransformationService.java
clinique: Reverted

J-N-K (dismissed):

J-N-K (approved):

PR #519, opened 2023-08-28
[amazonechocontrol] Fix lastVoiceCommand channel

J-N-K:
After an API change the lastVoiceCommand channel was no longer working. This introduces manual polling after a voice command was detected. Unfortunately this results in a delay of about 10-15s (compared to the old solution).


J-N-K:
This did not work reliably

PR #520, opened 2023-08-29
[amazonechocontrol] Fix lastVoiceCommand channel

J-N-K:
After an API change the lastVoiceCommand channel was no longer working. This introduces manual polling after a voice command was detected. Unfortunately this results in a delay of about 10-15s (compared to the old solution).


J-N-K:
This did not work reliably

jimSelf1:
First of all, thanks for your 3rd party echo binding! It brought back some things for me such as announcements and tts!
Now I saw this entry, does that mean, the lastVoiceCommand still isn't possible? Seems like the same is true for lastSpokenText.
I am using the 4.2.0 version of SmartHome/J Amazon Echo Control Binding.

PR #531, opened 2023-10-29
[basicprofiles] State filter profile

seime:
Simple filter profile to filter unwanted values originating from a Handler - based on some item(s)' state.

Optional support for replacement state to send instead

### DIFF
--- a/bundles/org.smarthomej.transform.basicprofiles/README.md
+++ b/bundles/org.smarthomej.transform.basicprofiles/README.md
@@ -192,7 +192,7 @@ Use case: Ignore values from a binding unless some other item(s) have a specific
 | Configuration Parameter | Type | Description                                                                                                                                                               |
 |-------------------------|------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `conditions`            | text | Comma separated list of expressions on the format `ITEM_NAME OPERATOR ITEM_STATE`, ie `MyItem EQ OFF`. Use quotes around `ITEM_STATE` to treat value as string ie `'OFF`' |
-| `mismatchState`         | text | Optional state to pass instead if conditions are NOT met. Use quotes to treat as `StringType`                                                                             |
+| `mismatchState`         | text | Optional state to pass instead if conditions are NOT met. Use quotes to treat as `StringType`. Defaults to `UNDEF`                                                         |
 
 Possible values for token `OPERATOR` in `conditions`:
 
--- a/bundles/org.smarthomej.transform.basicprofiles/README.md
+++ b/bundles/org.smarthomej.transform.basicprofiles/README.md
@@ -18,9 +18,9 @@ The given Command value is parsed either to `IncreaseDecreaseType`, `NextPreviou
 
 ```java
 Switch lightsStatus {
-        channel="hue:0200:XXX:1:color",
+    channel="hue:0200:XXX:1:color",
     channel="deconz:switch:YYY:1:buttonevent" [profile="basic-profiles:generic-command", events="1002,1003", command="ON"]
-        }
+}
 ```
 
 ## Generic Toggle Switch Profile
@@ -37,9 +37,9 @@ The Generic Toggle Switch Profile is a specialization of the Generic Command Pro
 
 ```java
 Switch lightsStatus {
-        channel="hue:0200:XXX:1:color",
+    channel="hue:0200:XXX:1:color",
     channel="deconz:switch:YYY:1:buttonevent" [profile="basic-profiles:toggle-switch", events="1002,1003"]
-        }
+}
 ```
 
 ## Debounce (Counting) Profile
@@ -87,7 +87,7 @@ It requires no specific configuration.
 
 The values of `QuantityType`, `PercentType` and `DecimalTypes` are negated (multiplied by `-1`).
 Otherwise the following mapping is used:
-
+ 
 `IncreaseDecreaseType`: `INCREASE` <-> `DECREASE`
 `NextPreviousType`: `NEXT` <-> `PREVIOUS`
 `OnOffType`: `ON` <-> `OFF`
@@ -173,15 +173,15 @@ Possible values for parameter `restoreValue`:
 
 ```Java
 Switch motionSensorFirstFloor {
-        channel="deconz:presencesensor:XXX:YYY:presence",
+    channel="deconz:presencesensor:XXX:YYY:presence",
     channel="deconz:colortemperaturelight:AAA:BBB:brightness" [profile="basic-profiles:time-range-command", inRangeValue=100, outOfRangeValue=15, start="08:00", end="23:00", restoreValue="PREVIOUS"]
-        }
+}
 ```
 
 ## State Filter Profile
 
 This filter passes on state updates from a (binding) handler to the item if and only if all listed item state conditions
-are met (conditions are ANDed togegheter).
+are met (conditions are ANDed together).
 Option to instead pass different state update in case the conditions are not met.
 State values may be quoted to treat as `StringType`.
 
@@ -204,5 +204,5 @@ Possible values for token `OPERATOR` in `conditions`:
 ```Java
 Number:Temperature airconTemperature{
         channel="mybinding:mything:mychannel"[profile="basic-profiles:state-filter",conditions="airconPower_item EQ ON",mismatchState="UNDEF"]
-        }
+}
 ```
--- a/bundles/org.smarthomej.transform.basicprofiles/README.md
+++ b/bundles/org.smarthomej.transform.basicprofiles/README.md
@@ -18,9 +18,9 @@ The given Command value is parsed either to `IncreaseDecreaseType`, `NextPreviou
 
 ```java
 Switch lightsStatus {
-    channel="hue:0200:XXX:1:color",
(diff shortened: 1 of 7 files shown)


bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/StateFilterProfile.java
```diff
@@ -0,0 +1,226 @@
...
+        this.itemRegistry = itemRegistry;
+
+        StateFilterProfileConfig config = context.getConfiguration().as(StateFilterProfileConfig.class);
+        if (config != null) {
+            conditions.addAll(parseConditions(config.conditions));
+            configMismatchState = parseState(config.mismatchState);
+        }
+    }
+
+    private List parseConditions(String config) {
```
J-N-K:
```suggestion
    private List parseConditions(@Nullable String config) {
```

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/StateFilterProfile.java
```diff
@@ -0,0 +1,226 @@
...
+        if (stateString == null) {
+            return null;
+        } else if (stateString.startsWith("'") && stateString.endsWith("'")) {
+            return new StringType(stateString.substring(1, stateString.length() - 1));
+        } else {
+            return TypeParser.parseState(acceptedDataTypes, stateString);
+        }
+    }
+
+    class Condition {
```
J-N-K:
```suggestion
    class StateCondition {
```

and move to a separate file

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/StateFilterProfile.java
```diff
@@ -0,0 +1,226 @@
...
+                this.value = value.substring(1, value.length() - 1);
+            }
+        }
+
+        public boolean matches(String state) {
+            switch (comparisonType) {
+                case EQ:
+                    if (!state.equals(value)) {
+                        return false;
+                    }
```
J-N-K:
```suggestion
                    return state.equals(value);
```

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/StateFilterProfile.java
```diff
@@ -0,0 +1,226 @@
...
+                    break;
+                case NEQ: {
+                    if (state.equals(value)) {
+                        return false;
+                    }
```
J-N-K:
```suggestion
                    return !state.equals(value);
```

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/StateFilterProfile.java
```diff
@@ -0,0 +1,226 @@
...
+
+        enum ComparisonType {
+            EQ,
+            NEQ
+        }
+
+        @Override
+        public String toString() {
+            return "Condition{" + "itemName='" + itemName + '\'' + ", comparisonType=" + comparisonType + ", value='"
+                    + value + '\'' + '}';
```
J-N-K:
```suggestion
            return "Condition{itemName='" + itemName + "', comparisonType=" + comparisonType + ", value='"
                    + value + "'}';
```

looks like the auto-generated `toString` from IntelliJ. I usually adjust that a bit.

bundles/org.smarthomej.transform.basicprofiles/README.md
```diff
@@ -177,3 +177,32 @@ Switch motionSensorFirstFloor {
...
+State values may be quoted to treat as `StringType`.
+
+Use case: Ignore values from a binding unless some other item(s) have a specific state.
+
+### Configuration
+
+| Configuration Parameter | Type | Description                                                                                                                                                               |
+|-------------------------|------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `conditions`            | text | Comma separated list of expressions on the format `ITEM_NAME OPERATOR ITEM_STATE`, ie `MyItem EQ OFF`. Use quotes around `ITEM_STATE` to treat value as string ie `'OFF`' |
+| `mismatchState`         | text | Optional state to pass instead if conditions are NOT met. Use quotes to treat as `StringType`                                                                             |
```
J-N-K: Why is it necessary to quote strings? There is also an issue with the way you split the string. If the comparison value contains a `,`, you split in the middle of a string. I know that this is really hard to handle correctly.

J-N-K (changes requested):

bundles/org.smarthomej.transform.basicprofiles/README.md
seime: Example why it may be useful (required) to quote strings

I have an aircon with a `mode` (string) channel. It takes values such as `HEAT`, `COOL`, `OFF`and so on. In order to not re-create state parsing, I've used the OH `TypeParser`. For a value `OFF` it will return `OnOffType.OFF`, and not the string `OFF`.

I guess I could improve further by checking the item type, but that would also make the same syntax mean 2 different things, and my gut feeling is that there will still be situations where the (hypotetical) user would want the non-default functionality.

Regarding comma as separator between multiple expressions; made this configurable.

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/StateFilterProfile.java
```diff
@@ -0,0 +1,219 @@
...
+    private final Logger logger = LoggerFactory.getLogger(StateFilterProfile.class);
+
+    private final ItemRegistry itemRegistry;
+    private final ProfileCallback callback;
+    private List<Class<? extends State>> acceptedDataTypes;
+
+    private List<StateCondition> conditions = new ArrayList<>();
+
+    @Nullable
+    private State configMismatchState = null;
```
J-N-K:
```suggestion
    private @Nullable State configMismatchState = null;
```

we inline null-annotations

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/StateFilterProfile.java
```diff
@@ -0,0 +1,219 @@
...
+@NonNullByDefault
+public class StateFilterProfile implements StateProfile {
+
+    private final Logger logger = LoggerFactory.getLogger(StateFilterProfile.class);
+
+    private final ItemRegistry itemRegistry;
+    private final ProfileCallback callback;
+    private List<Class<? extends State>> acceptedDataTypes;
+
+    private List<StateCondition> conditions = new ArrayList<>();
```
J-N-K:
```suggestion
    private List<StateCondition> conditions = List.of();
```

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/StateFilterProfile.java
```diff
@@ -0,0 +1,219 @@
...
+    private State configMismatchState = null;
+
+    public StateFilterProfile(ProfileCallback callback, ProfileContext context, ItemRegistry itemRegistry) {
+        this.callback = callback;
+        acceptedDataTypes = context.getAcceptedDataTypes();
+        this.itemRegistry = itemRegistry;
+
+        StateFilterProfileConfig config = context.getConfiguration().as(StateFilterProfileConfig.class);
+        if (config != null) {
+            conditions.addAll(parseConditions(config.conditions, config.separator));
```
J-N-K:
```suggestion
            conditions = parseConditions(config.conditions, config.separator);
```

bundles/org.smarthomej.transform.basicprofiles/src/main/java/org/smarthomej/transform/basicprofiles/internal/profiles/StateFilterProfile.java
```diff
@@ -0,0 +1,219 @@
...
+            configMismatchState = parseState(config.mismatchState);
+        }
+    }
+
+    private List<StateCondition> parseConditions(@Nullable String config, String separator) {
+        if (config == null)
+            return List.of();
```
J-N-K:
```suggestion
        if (config == null) {
            return List.of();
        }
```

always use braces

J-N-K (changes requested):
Just a few code-style comments, otherwise fine. Can you address those quickly? I would like to include that in the release later this week.

PR #542, opened 2023-12-21
[tuya] Add dimmer reversed value support

jsetton:
Some dimmer type channel should have the ability to support inverted values. This is useful for the `temp_value` channel which is generally implemented, compared to similar light control addons, from colder (0%) to warmer (100%).

### DIFF
--- a/bundles/org.smarthomej.binding.tuya/src/main/java/org/smarthomej/binding/tuya/internal/config/ChannelConfiguration.java
+++ b/bundles/org.smarthomej.binding.tuya/src/main/java/org/smarthomej/binding/tuya/internal/config/ChannelConfiguration.java
@@ -25,6 +25,7 @@ public class ChannelConfiguration {
     public int dp2 = 0;
     public int min = Integer.MIN_VALUE;
     public int max = Integer.MAX_VALUE;
+    public boolean reversed = false;
     public String range = "";
     public String irCode = "";
     public int irSendDelay = 300;
--- a/bundles/org.smarthomej.binding.tuya/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.smarthomej.binding.tuya/src/main/resources/OH-INF/thing/thing-types.xml
@@ -149,6 +149,12 @@
 			<parameter name="max" type="integer">
 				<label>Maximum</label>
 			</parameter>
+			<parameter name="reversed" type="boolean">
+				<label>Reversed</label>
+				<description>Changes the direction of the scale (e.g. 0 becomes 100, 100 becomes 0).</description>
+				<default>false</default>
+				<advanced>true</advanced>
+			</parameter>
 		</config-description>
 	</channel-type>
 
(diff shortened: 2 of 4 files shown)


J-N-K:
Wouldn't this also be possible with the INVERT profile?

jsetton:
As far as I can see (using the latest OH `4.2-SNAPSHOT`), I don't see any INVERT profile available. However, I am aware that a MAP or a SCRIPT profile can achieve this. But I think this is added complexity especially since the channel definitions for the Tuya addon are already generic, I figured it wouldn't hurt to have that option at that level.

I could make the case that the addon should actually invert the `temp_value` data point by default to keep the consistency across the OH ecosystem. If you rather hardcode that use case, I don't have any issue with that.

bundles/org.smarthomej.binding.tuya/src/main/java/org/smarthomej/binding/tuya/internal/config/ChannelConfiguration.java
```diff
@@ -25,6 +25,7 @@ public class ChannelConfiguration {
     public int dp2 = 0;
     public int min = Integer.MIN_VALUE;
     public int max = Integer.MAX_VALUE;
+    public Boolean inverted = Boolean.FALSE;
```
J-N-K:
```suggestion
    public boolean reversed = false;
```

Use primitives where possible. Since a default is provided, a primitive is sufficient here. I would also suggest to name the parameter "reversed", because an inversion (and that was my fault when I suggested the profile) us usually associated with a 100 -> -100 transformation.

bundles/org.smarthomej.binding.tuya/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -149,6 +149,11 @@
 			<parameter name="max" type="integer">
 				<label>Maximum</label>
 			</parameter>
+			<parameter name="inverted" type="boolean">
+				<label>Inverted</label>
```
J-N-K:
```suggestion
			<parameter name="reversed" type="boolean">
				<label>Reversed</label>
				<description>Changes the direction of the scale (e.g. 0 becomes 100, 100 becomes 0).</description>
```

J-N-K (changes requested):
Otherwise LGTM. Since this is also backward compatible we can merge it without any further issues.

jsetton:
I made the changes based on your comments.

J-N-K:
Looks good. the only thing missing is the documentation. Can you add the parameter to the channel documentation in README.md?

jsetton:
I updated the documentation.

J-N-K (dismissed):
Thanks for the other doc improvements!

J-N-K:
Unfortunately there is now a merge conflict after I merged your other PR. Can you fix that? Thanks.

jsetton:
Done.

J-N-K (approved):

PR #563, opened 2024-01-28
[tuya] Fix compiler warnings and simplify code

J-N-K:


J-N-K:
@d51x Can you have a look here? I fixed some code in the IR part.

d51x:
> @d51x Can you have a look here? I fixed some code in the IR part.

it looks ok, but needs to be tested

PR #592, opened 2024-10-04
[amazonechocontrol] Broken last voice command on Fire TV Cube

Trinitus01:
fixed: boken last voice command on Fire TV Cube


bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/handler/AccountHandler.java
```diff
@@ -141,6 +144,8 @@ public class AccountHandler extends BaseBridgeHandler implements PushConnection.
     private boolean disposing = false;
     private @Nullable AccountTO accountInformation;
 
+    Set<String> deviceSerialNumbers = Collections.synchronizedSet(new LinkedHashSet<>());
```
J-N-K: Why don't you use `ConcurrentHashMap.newKeySet()`? That's what we usually do if we want to create a new thread-safe `Set`.

J-N-K (commented):
Otherwise LGTM

bundles/org.smarthomej.binding.amazonechocontrol/src/main/java/org/smarthomej/binding/amazonechocontrol/internal/handler/AccountHandler.java
J-N-K: @Trinitus01 The "final" solution `CopyOnWriteArraySet` does not work, because it's iterator does not support remove and throws an `UnsupportedOperationException`. Is there any reason why this is a class field and not a local variable? As far as I can see that would also be possible.

Trinitus01: It seems there is no reason why this is a class field. So we can change it to a local one.

PR #593, opened 2024-10-04
[viessmann] fix IllegalStateException while parsing JSON

rogrun:
Viessmann changed the JSON response and added spaces and CR/LF. This PR removes these so that some needed values ​​can be searched and replaced.

This error was reported via email.


J-N-K (approved):

J-N-K:
Build succeeds locally.

PR #621, opened 2025-04-25
Correct sendAsString typo

mjagdis:
The parameter name in the resource file is "sendAdString" instead of "sendAsString".

However... given that it's been wrong for a year, defaults to off, couldn't be changed, and no one has complained maybe the "sendAsString" support should just be removed?


J-N-K (approved):

J-N-K:
Can you please port this to openhab-addons? For openHAB v5 Tuya will be removed from smarthomej

Issue #609, opened 2025-01-24
Tuya DPs not refreshing


krishgcek:
I am using Ver 4.2.x of the addon. The plugin in general works, I am able to control the plug, but the voltage, current etc... are not updated. 
With tiny tuya, everything works.
I have followed this e.g https://github.com/jasonacox/tinytuya/blob/d7e621f1ccfb993eb1edb8e9ff044b748dc7a55b/examples/getstatus.py
These lines has to be enabled.
 payload = d.generate_payload(tinytuya.UPDATEDPS,['18','19','20'])
 d.send(payload)

I think the problem is here,
 public void refreshStatus() {
        MessageWrapper<?> m = new MessageWrapper<>(DP_REFRESH, Map.of("dpId", List.of(4, 5, 6, 18, 19, 20)));
        Channel channel = this.channel;
        if (channel != null) {
            channel.writeAndFlush(m);
        } else {
            logger.warn("{}: Refreshing status failed. Device is not connected.", deviceId);
        }
    }

When I add 4,5,6 with python I got empty result.

Can you make the DPs to be queried configurable?

BigGeorgeTx:
In Ver 4.3.1 (and for quite a while with previous versions), I am not getting any Item state updates from the Tuya binding.  As stated by @krishgcek , control commands are executing as expected.  But the states for all of my Tuya devices are never updated for activity triggered elsewhere (either through the SmartHome app or someone manually flipping as switch).  (As recommended, if I use the SmartHome app, I force it closed when I’m finished as advised).

PR #622, opened 2025-04-26
Poll with DP_QUERY instead of DP_REFRESH

mjagdis:
Should polling use DP_QUERY to request current DP values rather than DP_REFRESH?

The only Tuya devices I have available are some (identical) smart plugs using the 3.5 protocol (for which I'm using the @CyborgAndy patches [here](https://github.com/CyborgAndy/smarthomej-addons.git)). However...

Using refreshStatus, which sends a DP_REFRESH, I get a response payload of "query dp failed". Doing a requestStatus, which sends a DP_QUERY, gets the full set of DPs as expected.

I have a suspicion that a DP_REFRESH is intended to trigger underlying sampling and return the new DP values and that "query dp failed" means no requested DPs were triggerable. DP_QUERY, on the other hand, is just a request for current values of DPs.

On the gripping hand these my plugs seem to do continuous sampling and the values returned for a DP_QUERY do update.Other devices, battery devices in particular, may only sample on request and need to be polled with DP_REFRESH? I have no way of testing this :-(

N.B. Issue #609 seems to imply that some smart plugs can be DP_REFRESHed for some DPs but the DP_REFRESH will fail if non-existent DPs are requested. My smart plugs fail DP_REFRESH for ANY set of DPs.

So...
 * We definitely need to do a DP_QUERY for some devices
 * We may need to do a DP_REFRESH for some devices
 * If we do a DP_REFRESH we should only request DPs known to exist (or some subset?)

As I said, I have nothing to test further with.

For the record the smart plugs I have are Eightree dual band (2.4GHz/5GHz) UK version (yes, they are advertised as 5GHZ but they are actually dual band)

### DIFF
--- a/bundles/org.smarthomej.binding.tuya/src/main/java/org/smarthomej/binding/tuya/internal/handler/TuyaDeviceHandler.java
+++ b/bundles/org.smarthomej.binding.tuya/src/main/java/org/smarthomej/binding/tuya/internal/handler/TuyaDeviceHandler.java
@@ -31,6 +31,7 @@ import java.util.Objects;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


mjagdis:
I've updated this to do a DP_REFRESH for DPs that are known to exist followed immediately by a DP_QUERY. This ought to work for everything although it's possible that there could be non-refreshable DPs that cause DP_REFRESH to fail and not trigger refresh of other DPs that need to be refreshed. Someone with a good selection of battery devices would need to experiment a bit. It's also possible that the DP_QUERY follows the DP_REFRESH too closely and might get values before the refresh completes and updates them. I don't _think_ that's a problem because if refresh does anything it should trigger reports of new values once they are available. If there are devices that need refresh but do not report then we might need to trigger the DP_QUERY from the "query dp fail" response (similar to how DP_QUERY_NO_SUPPORT works) but again, someone with battery devices...

J-N-K (approved):
LGTM

J-N-K:
Please also add this to openhab-addons

J-N-K:
Unfortunately I can't merge the 3.5 protocol changes since the PR was closed.

PR #623, opened 2025-04-28
Fix to allow for Channels whose DPs have string values

mjagdis:
As title. DPs with string values didn't get Channels created for them because SchemaDp didn't know what to do with them.


J-N-K (approved):
LGTM

PR #624, opened 2025-04-29
Get channel labels from schema label or map from schema code

mjagdis:
Allow schemas to provide labels for Channels and if none is provided map the code to something reasonably presentable.

It may seem like Channel labels are something of an internal detail in openHAB but they are used in generating the default names of linked Items and thence to places likely to be exposed to users.


J-N-K:
Is this backward compatible or do we need to delete all things and re-add them?

mjagdis:
Existing Things will retain their current Channel labels, newly created Things will get the new Channel labels. The only issue is that labels (both Thing and Channel) are used to construct the names of Items so any linked Items may be left with "confusing" names. Sadly Items names are definitive identifiers (there are no Item UIDs) and are used in Channel<->Item links, persistence etc. But that's really a core openHAB problem :-(

J-N-K (approved):
LGTM

